PR #1, opened 2019-04-11
Sketch out the basic structure.

ltratt:
This is currently capable of running hello world and nothing beyond that. Even then, it cheats in a few ways. However, this is much more about getting a minimalish skeleton of a VM implemented that we can start expanding and correcting.

The key component is the memory design: we have a reference counted GC that exposes trait objects as single machine word big pointers and that only requires a single malloc per object. Getting to this point has involved a number of false starts: the eventual design is simple and, I hope, relatively non-clunky to use. In essence, the `Gc` type in `gc.rs` is roughly similar to `Rc` but: you can turn a `Gc` into a pointer and back into a cloned-and-incremented `Gc` in one go; and using `GcBox::alloc_blank` you can allocate a `Gc` of arbitrary size (which is very handy when you're creating trait objects). The `ThinObj` type in `vm/objects.rs` then stores away vtable pointers so that we end up with thin pointers. The most annoying hack is the `GcLayout` trait, and even that's not *too* bad. More to the point, the design should hopefully allow us to move fairly easily to a proper GC in the future.

In terms of the more yksom-specific things, the key is probably in `vm/objects.rs`. The `Val` struct hides whether it is storing a boxed object or an unboxed integer (or so on). Boxed objects implement the `Obj` trait. The key function is then `Val::gc_obj` which returns a boxed object (boxing as necessary). Structuring things in this way makes it relatively easy to use things with the borrow checker. `Obj`s have a method `as_any` which casts them to `Any`: this allows you to take an `Obj`, convert it to an `Any`, and then use `downcast_ref` to safely dynamically get at the particular struct (`Class`, `String_` and so on) inside it.

Currently we only implement a miniscule (and not correct -- println should be on System, not on String) subset of SOM's standard library in `lib/SOM`. This is enough to run hello world:

```
Hello = (
    run = ('Hello, World from SOM' println )
)
```

One nice touch relative to other SOMs is that we get proper parsing error messages through lrpar (and a less broken parser than SOM).

Anyway, there's still a lot to do here, and I'm sure that some parts of this design will prove to be bone-headed. Which, of course, remains to be seen...

Because this is such early stages stuff, you should expect to see lots of less than perfect things in this PR, and we can't expect everything to be fixed at this point. But it would be good to spot flaws that could hurt us in the long-term! @jacob-hughes I'd suggest you want to have a good gander at `gc.rs` and the associated parts in `vm/objects.rs`. @ptersilie You probably want to check the grammar and the (very basic compiler. @vext01 You probably want to check pretty much every other part.


src/lib/compiler/som.y
```diff
@@ -0,0 +1,201 @@
...
+ClassDef -> Result<Class, ()>:
+      "ID" "=" SuperClass "(" InstanceFields MethodsOpt ClassMethods ")"
+      { Ok(Class{ name: map_err($1)?, supername: $3?, methods: $6? }) }
+    ;
+SuperClass -> Result<Option<Lexeme<StorageT>>, ()>:
+      "ID" { Ok(Some(map_err($1)?)) }
+    | { Ok(None) }
+    ;
+InstanceFields -> Result<(), ()>:
+      "|" InstanceFieldNames "|" { unreachable!() }
```
ptersilie: Why is this `unreachable`? It looks like I can reach it via `InstanceFields` from within `ClassDef`.

ltratt: I didn't know about the `unimplemented!()` macro when I did that bit ;) I'll change it any any others in the grammar.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -0,0 +1,190 @@
...
+        match self.sends.entry(m) {
+            Entry::Occupied(e) => *e.get(),
+            Entry::Vacant(e) => {
+                e.insert(off);
+                off
+            }
+        }
+    }
+
+    fn c_method(
```
vext01: So is the `c` for compile?

src/lib/compiler/som.y
ptersilie: Fair. I assumed something like that, but just wanted to make sure.

ptersilie:
Looks good to me so far.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    fn c_body(
+        &mut self,
+        name: (Lexeme<StorageT>, &str),
+        body: &ast::MethodBody,
+    ) -> Result<cobjects::MethodBody, Vec<(Lexeme<StorageT>, String)>> {
+        match body {
+            ast::MethodBody::Primitive => match name.1 {
+                "new" => Ok(cobjects::MethodBody::Primitive(Primitive::New)),
```
vext01: I'm surprised that a `print` or a `new` is considered a whole method body. Wouldn't they be better modelled as a statement? Does this mean that (e.g.) `print` can only appear in a function body on its own? Or did I misunderstand?

src/lib/vm/objects.rs
```diff
@@ -0,0 +1,522 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct Class {
+    pub path: PathBuf,
+    pub methods: HashMap<String, Method>,
+    pub instrs: Vec<Instr>,
+    pub consts: Vec<Val>,
+    pub sends: Vec<(String, usize)>,
```
vext01: What's a `send` here and how does it relate to a method, which we already have a member for.

src/lib/vm/objects.rs
```diff
@@ -0,0 +1,522 @@
...
+        Err(())
+    }
+
+    fn get_class(&self, vm: &VM) -> Val {
+        vm.cls_cls.clone()
+    }
+}
+
+impl Class {
+    pub fn from_ccls(vm: &VM, ccls: cobjects::Class) -> Self {
```
vext01: ccls?

compiled class?

src/lib/vm/objects.rs
```diff
@@ -0,0 +1,522 @@
...
+        unimplemented!();
+    }
+}
+
+#[derive(Debug)]
+pub struct Inst {
+    class: Val,
+}
+
+impl Obj for Inst {
```
vext01: What's an `Inst`? A bytecode instruction?

Are instructions first class in SOM then?

It's more likely that I have no idea what I'm talking about ;)

src/lib/compiler/ast_to_instrs.rs
ltratt: Yes.

src/lib/compiler/ast_to_instrs.rs
ltratt: That's Smalltalk/SOM's semantics. You basically say things like:

```
new = primitive
```

and calling `new` then semi-magically calls the primitive operation (also called `new`) in the VM.

src/lib/vm/objects.rs
ltratt: Message send. I'm actually a bit fuzzy on the precise semantics of this in SOM, so I'm not sure if I've got that right yet.

src/lib/vm/objects.rs
ltratt: Yes.

src/lib/vm/objects.rs
ltratt: Instance.

vext01: Woops! Comment and/or clearer name needed!

ltratt: OK.

src/lib/vm/objects.rs
vext01: I always thought a send was a method invocation. I'm not sure why we'd store such a thing. Maybe we will find out later then.

src/lib/compiler/ast_to_instrs.rs
vext01: OK, I'm not very clear on this, but I trust you!

src/lib/vm/objects.rs
ltratt: A send is a method invocation -- sort of. In most languages `o.f()` is equivalent to `let x = o.f; x();` but in SOM it means "ask `o` to do something with message `f`". Because the sends are strings (it's dynamically typed), I'm trying to optimise the common case of a module calling a function `f` repeatedly (i.e. I only store `f`'s name once).

vext01: OK

[I think I need to read up on SOM!]

src/lib/vm/gc.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub fn alloc_blank(l: Layout) -> (*mut Self, *mut T) {
+        let (layout, valoff) = Layout::new::<GcBox<T>>().extend(l).unwrap();
+        debug_assert_eq!(valoff, size_of::<GcBox<T>>());
+        let gcbptr = unsafe { alloc(layout) as *mut Self };
+        if gcbptr.is_null() {
+            panic!("Can't allocate memory.");
+        }
+        unsafe { &mut *gcbptr }.clones = 1;
+        let valptr = unsafe { (gcbptr as *mut u8).add(valoff) } as *mut T;
+        (gcbptr, valptr)
```
jacob-hughes: Let's see if I understand this correctly. `alloc_blank` returns two pointers: a box pointer to the `GcBox` object (which contains the ref count); and a value pointer to the underlying object. 

Assuming I've got that correct, am I right in thinking that the value pointer will _always_ point to the address in memory which comes directly after the fields in a `GcBox`? If so, why do we return two pointers and hide the fact that second is derived from some fixed offset?

src/lib/vm/gc.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+impl<T: GcLayout> Deref for GcBox<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        let valptr = unsafe { (self as *const GcBox<T> as *const u8).add(size_of::<GcBox<T>>()) }
```
jacob-hughes: I'm not sure I fully understand this part. To deref the object, we need to do some pointer arithmetic as the object is always `*const GcBox + size_of(GcBox)`.

What does  ```(self as *const GcBox<T> as *const u8)``` mean? Why do we cast the pointer to a `u8` before the arithmetic?

jacob-hughes: Also, should we be using `offset()` for arithmetic? https://doc.rust-lang.org/std/primitive.pointer.html#method.offset

src/lib/vm/gc.rs
```diff
@@ -0,0 +1,113 @@
...
+impl<T: GcLayout> Deref for Gc<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        unsafe { &*self.ptr }.deref()
+    }
+}
+
+#[derive(Debug)]
+pub struct GcBox<T: ?Sized> {
```
jacob-hughes: Why the decision not to make the actual Gc object a field inside the box? Is this so that you can allocate an object of arbitrary size?

src/lib/vm/gc.rs
ltratt: Yes, you're correct but there's no reason why the two pointers have to be next to each other in memory. The first pointer could be to some private GC structure stored elsewhere and the second pointer to a freshly malloced block. Splitting them apart seemed to me to give a bit more flexibility to `Gc`/`GcBox`. It's not a big deal though.

src/lib/vm/gc.rs
ltratt: Yes, you're right (in your first point).  `add` and `offset` are more-or-less the same except `add` does so in bytes. I ended up finding it less confusing to use `add` everywhere, hence the cast to `*const u8`.

src/lib/vm/gc.rs
ltratt: Yes, it's because then we can have unsized objects without that complexity leaking outside quite as obviously.

ltratt:
I've added a commit with some simplifications suggested by Clippy. @jacob-hughes I guess you still have some follow-up comments based on my comments?

src/lib/vm/gc.rs
jacob-hughes: > but there's no reason why the two pointers have to be next to each other in memory

Is this true? I'm not sure how de-referencing would then be possible if it's just offsetting based on the `GcBox` size.

src/lib/vm/gc.rs
jacob-hughes: That makes sense

src/lib/vm/gc.rs
jacob-hughes: That sounds reasonable to me

src/lib/vm/gc.rs
ltratt: `Gc::deref` can do whatever it wants, so it can e.g. follow multiple pointers if it wants to. I'm not saying that's a good idea ;) but it is possible.

ltratt:
@jacob-hughes Assuming you're now happy, I think this is ready to be squashed?

jacob-hughes:
LGTM

ltratt:
Squashed.

PR #2, opened 2019-04-13
Drop memory properly

ltratt:
Previously we didn't drop memory in many cases. Running `valgrind` on hello world shows this clearly:

```
==3110== HEAP SUMMARY:
==3110==     in use at exit: 8,389 bytes in 44 blocks
==3110==   total heap usage: 27,864 allocs, 27,820 frees, 17,935,698 bytes allocated
==3110==
==3110== LEAK SUMMARY:
==3110==    definitely lost: 0 bytes in 0 blocks
==3110==    indirectly lost: 0 bytes in 0 blocks
==3110==      possibly lost: 0 bytes in 0 blocks
==3110==    still reachable: 8,389 bytes in 44 blocks
==3110==         suppressed: 0 bytes in 0 blocks
```

This PR fixes that.

First, some of those leaks are because we called `process::exit` without giving parent frames the chance to `drop` items. This is fixed in https://github.com/softdevteam/yksom/commit/1cc668dbf3e719d4167c0329cb293101453c20db.

Second, `Gc`/`GcBox`/`ThinObj` didn't call `drop` at all, mostly because I wasn't really sure how to do this on trait objects. `Box::drop` has this wonderful comment:

```
// FIXME: Do nothing, drop is currently performed by compiler.
```

Fortunately `ptr::drop_in_place` does what we want. https://github.com/softdevteam/yksom/commit/f272e3f5c2e50ce4c1cf10453b9734263e113e03 has the relevant fixes.

With this, we no longer leak any memory on hello world:

```
==2881== HEAP SUMMARY:
==2881==     in use at exit: 48 bytes in 2 blocks
==2881==   total heap usage: 27,864 allocs, 27,862 frees, 17,935,698 bytes allocated
==2881==
==2881== LEAK SUMMARY:
==2881==    definitely lost: 0 bytes in 0 blocks
==2881==    indirectly lost: 0 bytes in 0 blocks
==2881==      possibly lost: 0 bytes in 0 blocks
==2881==    still reachable: 48 bytes in 2 blocks
==2881==         suppressed: 0 bytes in 0 blocks
```

The 2 blocks that show up in `valgrind` are to do with mutexes, so we're not responsible for them.

### DIFF
--- a/src/lib/vm/gc.rs
+++ b/src/lib/vm/gc.rs
@@ -111,9 +111,11 @@ impl<T: GcLayout> Deref for GcBox<T> {
     type Target = T;
 
     fn deref(&self) -> &T {
-        let valptr = unsafe { (self as *const GcBox<T> as *const u8).add(size_of::<GcBox<T>>()) }
-            as *const T;
-        unsafe { &*valptr }
+        unsafe {
+            let valptr =
+                (self as *const GcBox<T> as *const u8).add(size_of::<GcBox<T>>()) as *const T;
+            &*valptr
+        }
     }
 }
 
--- a/src/lib/vm/gc.rs
+++ b/src/lib/vm/gc.rs
@@ -12,6 +12,7 @@ use std::{
     marker::PhantomData,
     mem::{forget, size_of},
     ops::Deref,
+    ptr,
 };
 
 /// Since Rust's alloc system requires us to tell it the `Layout` of a region of memory upon
@@ -64,7 +65,10 @@ impl<T: GcLayout> Drop for Gc<T> {
     fn drop(&mut self) {
         let clones = unsafe { &mut *self.ptr }.clones;
         if clones == 1 {
-            unsafe { dealloc(self.ptr as *mut u8, self.layout()) };
+            unsafe {
+                ptr::drop_in_place(self.ptr);
+                dealloc(self.ptr as *mut u8, self.layout());
+            }
         } else {
             unsafe { &mut *self.ptr }.clones = clones - 1;
         }
@@ -80,7 +84,7 @@ impl<T: GcLayout> Deref for Gc<T> {
 }
 
 #[derive(Debug)]
-pub struct GcBox<T: ?Sized> {
+pub struct GcBox<T> {
     clones: usize,
     phantom: PhantomData<T>,
 }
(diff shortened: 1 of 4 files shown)


src/lib/vm/gc.rs
```diff
@@ -64,7 +65,10 @@ impl<T: GcLayout> Drop for Gc<T> {
     fn drop(&mut self) {
         let clones = unsafe { &mut *self.ptr }.clones;
         if clones == 1 {
-            unsafe { dealloc(self.ptr as *mut u8, self.layout()) };
+            unsafe {
+                ptr::drop_in_place(self.ptr);
```
jacob-hughes: So, if I understand correctly we're: (a) derefing the pointer to call the object's `drop` method; and (b) explicitly freeing the memory and allowing the allocator to re-use it (`dealloc` call).

ltratt: More or less. `drop_in_place` is magic (I assume that, if `T` is a trait object, it has a special slot for `drop`), but it is reasonable to assume that it derefs `T` and calls `drop`.

[We'll need support for this in your GC too I think!]

src/lib/vm/gc.rs
```diff
@@ -80,7 +84,7 @@ impl<T: GcLayout> Deref for Gc<T> {
 }
 
 #[derive(Debug)]
-pub struct GcBox<T: ?Sized> {
```
jacob-hughes: Why do we now disallow DSTs here? I can't see why your `drop` implementation wouldn't work if the size of `T` wasn't known statically (but I'm guessing there's probably an example of where this breaks).

src/lib/vm/gc.rs
jacob-hughes: :+1:  I understand

src/lib/vm/objects.rs
```diff
@@ -243,6 +244,16 @@ impl Deref for ThinObj {
     }
 }
 
+impl Drop for ThinObj {
+    fn drop(&mut self) {
+        let self_ptr = self as *const Self;
+        unsafe {
+            let obj_ptr = self_ptr.add(1);
+            ptr::drop_in_place::<Obj>(transmute((obj_ptr, self.vtable)));
```
jacob-hughes: Again, I'll check that I understand correctly what's happening here: A thin object is comprised of a `vptr` followed immediately by an `obj_ptr`. 

It looks like here we're turning our `ThinObj` pointer back into the fat pointer which Rust understands so that `drop_in_place` works.

ltratt: Correct.

src/lib/vm/gc.rs
ltratt: From memory, if this is left in, `Drop` then has to deal with unsized things, and then lots of things get messy. Given that I don't need to deal with unsized things in yksom, the mess didn't seem worth it...

jacob-hughes: Fair enough

jacob-hughes:
I'm happy with this, I don't really have any comments for improvements as this pushes the boundaries of what I know about Rust. But hopefully if I continue pressing you on _why_ certain design decisions were made we might stumble upon some design flaw ;)

ltratt:
Yes, keep pressing! I'm also at the edge of my understanding here.

jacob-hughes:
I'm guessing you don't need to squash? If so I'll merge

ltratt:
I think it can be merged as-is.

PR #3, opened 2019-04-18
Fix boxing of ints

ltratt:
There is an element of "2 for 1" in this PR. https://github.com/softdevteam/yksom/commit/23a03e051d3cdfc2505cc5fff7ae8f46d0642dd6 fixes an infinite recursion problem in boxing integers. While doing that, I noticed that we had become highly inconsistent in the return types we used, particularly in integer creating/boxing/unboxing.  https://github.com/softdevteam/yksom/commit/833ddb8fc562c06b15dfb6dca0c812c3f6420e60 fixes that problem.

### DIFF
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -91,22 +91,22 @@ impl Val {
     }
 
     /// Return this `Val`'s box. If the `Val` refers to an unboxed value, this will box it.
-    pub fn gc_obj(&self, vm: &VM) -> Gc<ThinObj> {
+    pub fn gc_obj(&self, vm: &VM) -> Result<Gc<ThinObj>, VMError> {
         match self.valkind() {
             ValKind::GCBOX => {
                 debug_assert_eq!(ValKind::GCBOX as usize, 0);
                 debug_assert_eq!(size_of::<*const GcBox<ThinObj>>(), size_of::<usize>());
                 debug_assert_ne!(self.val, 0);
-                unsafe { Gc::clone_from_raw(self.val as *const _) }
+                Ok(unsafe { Gc::clone_from_raw(self.val as *const _) })
             }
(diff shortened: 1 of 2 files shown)


vext01:
So "boxed" here is something "GCable"?

ltratt:
It's used with its normal meaning: "allocated on the heap".

vext01:
There seems to be some overloading at play. In interpreter-speak, a boxed integer (for example) is an integer in a wrapper, which could live anywhere. But in Rust-speak, something boxed is on the heap. I think the latter is the intended meaning, right?

I wonder if there's a way to clarify, because I assumed the wrong meaning.

vext01:
[last query resolved in mattermost]

> Previously we cause an infinite loop if we tried to get the boxed value of an
unboxed int

So is getting a boxed value from an unboxed int an error case, or do you make a box?

ltratt:
>So is getting a boxed value from an unboxed int an error case, or do you make a box?

`gc_obj` boxes unboxed ints.

src/lib/vm/objects.rs
```diff
@@ -440,10 +441,9 @@ impl Int {
         } else if i & (1 << (BITSIZE - 1)) == 0 {
             // One of the top TAG_BITSIZE bits is set, but not the topmost bit itself, so we can
             // box this as an isize.
-            Ok(Int::from_isize(vm, i as isize))
+            Ok(Int::from_isize(vm, i as isize)?)
```
vext01: So in theory this `?` can never fail?

ltratt: That depends entirely on `from_isize`. Even if it might not fail now, a change in the future might make it fallible, hence why we need to percolate `VMError` more-or-less everywhere.

vext01:
I think this looks OK, but let me be upfront that I have no idea of the language semantics you are implementing!

ltratt:
It's OK, SOM has no published semantics! It's just an implementation. So I don't know what I'm doing either.

PR #4, opened 2019-04-18
Allow object references to be converted back into ThinCells/Gcs.

ltratt:
Without this we can't write VM functions that return `&self` as a `Val`, because `&self` will be a reference to a concrete struct, and we can't turn that into a Val, because `Val::from_obj` takes a `T`, not an `&T`. At its simplest, this means we couldn't write the following identity function:

```
  struct S;

  impl Obj for S {}

  impl S {
      fn identity(&self) -> Val {
          // what goes here?
      }
  }
```

This commit adds a way to turn a `&T` back into a ThinObj/GcBox/Gc. It allows us to express the identity function as follows:

```
  impl S {
      fn identity(&self) -> Val {
          Val::restore(self)
      }
  }
```

This is clearly dangerous, because it assumes that the memory pointed to by `&self` is part of a ThinObj/GcBox combo. If that isn't true, all sorts of bad things will happen. Fortunately it's relatively easy to obey this constraint by always passing new instances immediately to `Val::from_obj`.

[I spent considerable time working out if I could enforce this constraint in the type system. My conclusion is that I probably could, but at the cost of making the code hugely unergonomic.]

### DIFF
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -322,7 +322,7 @@ impl Obj for Class {
 }
 
 impl Class {
-    pub fn from_ccls(vm: &VM, ccls: cobjects::Class) -> Self {
+    pub fn from_ccls(vm: &VM, ccls: cobjects::Class) -> Val {
         let mut methods = HashMap::with_capacity(ccls.methods.len());
         for cmeth in ccls.methods.into_iter() {
             let body = match cmeth.body {
@@ -342,13 +342,16 @@ impl Class {
                 cobjects::Const::String(s) => String_::new(vm, s),
             })
             .collect();
-        Class {
-            path: ccls.path,
-            methods,
-            instrs: ccls.instrs,
-            consts,
-            sends: ccls.sends,
-        }
+        Val::from_obj(
+            vm,
+            Class {
+                path: ccls.path,
+                methods,
+                instrs: ccls.instrs,
+                consts,
+                sends: ccls.sends,
+            },
+        )
     }
 
     pub fn get_method(&self, _: &VM, msg: &str) -> Result<&Method, VMError> {
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -9,6 +9,28 @@
 
 #![allow(clippy::new_ret_no_self)]
 
+//! This file contains the core SOM objects. Note that there is a fundamental constraint that
+//! *must* be obeyed by the programmer at all times: upon their creation, instances of the `Obj`
+//! trait must immediately be passed to `Val::from_obj`. In other words this is safe:
+//!
+//! ```rust,ignore
+//! let x = Val::from_obj(vm, String_{ s: "a".to_owned() });
+//! dbg!(x.gc_obj().as_str());
+//! ```
+//!
+//! but this leads to undefined behaviour:
+//!
+//! ```rust,ignore
+//! let x = String_{ s: "a".to_owned() };
+//! dbg!(x.gc_obj().as_str());
+//! ```
+//!
+//! The reason for this is that methods on `Obj`s can call `Val::restore` which converts an `Obj`
+//! reference back into a `Val`.
+//!
+//! Although this constraint is not enforced through the type system, it is not hard to obey: as
+//! soon as you create an `Obj` instance, pass it to `Val::from_obj`.
+
 use std::{
     alloc::Layout,
     any::Any,
@@ -73,6 +95,18 @@ impl Val {
         let ptr = ThinObj::new(obj).into_raw();
         Val {
             val: unsafe { transmute(ptr) },
+        }
+    }
+
+    /// Convert `obj` into a `Val`. `Obj` must previously have been created via `Obj::from_off` and
+    /// then turned into an actual object with `gc_obj`: failure to follow these steps results in
+    /// undefined behaviour.
+    pub fn recover(_: &VM, obj: &Obj) -> Self {
+        unsafe {
+            let ptr = ThinObj::recover(obj);
+            Val {
+                val: transmute(ptr),
+            }
         }
     }
 
@@ -167,6 +201,7 @@ impl Drop for Val {
     }
 }
 
+/// The main SOM Object trait.
 pub trait Obj: Debug + GcLayout {
     /// Return this object as an `Any` that can then be safely turned into a specific struct.
     fn as_any(&self) -> &Any;
@@ -220,6 +255,11 @@ impl ThinObj {
         }
         forget(v);
         unsafe { Gc::from_raw(gcbptr) }
+    }
+
+    pub unsafe fn recover(o: &Obj) -> Gc<ThinObj> {
+        let thinptr = (o as *const _ as *const u8).sub(size_of::<ThinObj>()) as *const ThinObj;
+        Gc::recover(thinptr)
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/lib/vm/objects.rs
```diff
@@ -76,6 +98,18 @@ impl Val {
         }
     }
 
+    /// Convert `obj` into a `Val`. `Obj` must previously have been created via `Obj::from_off` and
+    /// then turned into an actual object with `gc_obj`: failure to follow these steps results in
+    /// undefined behaviour.
+    pub fn recover(_: &VM, obj: &Obj) -> Self {
```
jacob-hughes: Why do we take a `&VM` here as the first argument if it's never used?

ltratt: Because one day we might. [Basically, bitter experience suggests it's always a good idea to pass `vm` around. If you don't use it, the optimiser should remove its costs. If you do want it one day... bingo! it's there.]

jacob-hughes: Fair enough.

src/lib/vm/objects.rs
```diff
@@ -76,6 +98,18 @@ impl Val {
...
+        unsafe {
+            let ptr = ThinObj::recover(obj);
+            Val {
+                val: transmute(ptr),
```
jacob-hughes: Most usages of `transmute` that I've come across in the rust docs have type annotations, so I'm trying to make sense of what's going on here. I'm assuming we're taking a `ptr` which could be of type `Gc<T>` or `*const GcBox<T>` and converting it to a `usize`.

I think that we should add explicit annotations here to make that clear.

src/lib/vm/objects.rs
ltratt: Although, on reflection, the chances of this function ever needing access to the `VM` type are close to zero (since it's really just fiddling with stuff. Maybe it is best to get rid of this argument.

src/lib/vm/objects.rs
ltratt: `transmute` is often annotated because the type checker can't work out what to do. But sometimes it can :) Here it's transmuting a pointer (from `ThinObj::recover`) to a `usize` (`val`).

src/lib/vm/objects.rs
ltratt: The new commit gets rid of it.

src/lib/vm/objects.rs
ltratt: I realised that there's a good reason for statically typing `transmute` beyond mere convention: to make sure that we're getting the conversion we actually expect (on the basis that `transmute` will happily map between types that we might not have intended). https://github.com/softdevteam/yksom/pull/4/commits/da0938fae9e169c894024ee6c5e142227368c503 does just that... and it found a bug. So it was definitely worthwhile!

src/lib/vm/objects.rs
```diff
@@ -237,19 +279,20 @@ impl Deref for ThinObj {
...
 
 impl Drop for ThinObj {
     fn drop(&mut self) {
-        let self_ptr = self as *const Self;
+        let self_ptr = self as *const Self as *const u8;
         unsafe {
-            let obj_ptr = self_ptr.add(1);
-            ptr::drop_in_place::<Obj>(transmute((obj_ptr, self.vtable)));
+            let obj_ptr = self_ptr.add(size_of::<ThinObj>());
+            let fat_ptr = transmute::<(*const u8, usize), &mut dyn Obj>((obj_ptr, self.vtable));
```
jacob-hughes: Are you sure that this is not UB? According to the nomicon:

> Transmuting an & to &mut is UB
    Transmuting an & to &mut is always UB
    No you can't do it
    No you're not special

Does that apply to raw pointers too? I certainly hope not, but it's not clear to me.

ltratt: Since we're `Drop`ing the pointer at this point, we know that (modulo other bugs!) we're the only place referring to that pointer, so we're safe having a `mut` pointer.

In the general case, though, we're subject to the same restrictions as `UnsafeCell`, which complement the nomicon explanation above.

jacob-hughes: The explanation makes sense, but I think that we can remove the UB by casting `self_ptr` to a `*mut u8` instead of a `*const u8` on line 219. That would then satisfy the `transmute` rules.

ltratt: Ah, I see what you're getting at. It's not UB to cast `*const u8` to `*mut u8`, because pointers are not subject to the same rules as references: pointers can only cause undefined behaviour when you dereference them. Admittedly, this does raise the question of what the difference between `*const u8` and `*mut u8` are: are they just hints to the programmer or does LLVM make use of that information to optimise things? If the latter, then I don't think that changing things to:

```rust
let fat_ptr = transmute::<(*mut u8, usize), &mut dyn Obj>((obj_ptr as *mut u8, self.vtable));
```

would make any difference?

jacob-hughes: This information about pointers and references is definitely passed to LLVM, I'm not sure how it uses it to optimise the IR though. It may not make any difference in this case but I do prefer `*mut u8` from a readability perspective.

ltratt: Fair enough. The new commit does just that.

[I do now need to find out if `*const` and `*mut` have correctness implications!]

jacob-hughes:
This passes the build now. Do you want to squash? Or are you still checking to see if you've affected correctness by changing from `*const` to `*mut`?

src/lib/vm/objects.rs
ltratt: https://internals.rust-lang.org/t/what-is-the-real-difference-between-const-t-and-mut-t-raw-pointers/6127 is interesting. Summary: I think we're fine.

jacob-hughes: Cool, I'm satisfied with this

ltratt:
Squashed.

src/lib/vm/objects.rs
```diff
@@ -237,19 +279,21 @@ impl Deref for ThinObj {
...
+            let obj_ptr = self_ptr.add(size_of::<ThinObj>());
+            transmute::<(*const u8, usize), &dyn Obj>((obj_ptr, self.vtable))
         }
     }
 }
 
 impl Drop for ThinObj {
     fn drop(&mut self) {
-        let self_ptr = self as *const Self;
+        let self_ptr = self as *const Self as *const u8;
```
jacob-hughes: For consistency, shouldn't we be casting this to `*mut u8` here too?

ltratt: I don't think so. We're not mutating the memory that houses `self`; we modify the memory immediately after it (which, from `self`'s perspective, could be anywhere in the heap).

jacob-hughes: That is true, but it's challenged an assumption I've made about raw pointers: do we annotate as `*mut T` only if we mutate the pointer's value; or, do we annotate as `*mut T` because we got the address from a mutable reference (`&mut self` in this case). I was assuming the latter.

ltratt: TBH, I don't think it makes any difference in this case. What we know would be unsafe is to have `*mut self` and then mutate it: that would be UB because we still have `&mut self` active. But we don't do that either.

jacob-hughes: That is true. Okay you've convinced me

jacob-hughes:
LGTM. I'll merge

PR #5, opened 2019-04-23
Provide a nicer way of casting to a concrete Obj type.

ltratt:
There is some inherent ickiness here, because given a `Val`, you always have to call `gc_obj` in order to get a `ThinObj`. From Rust's perspective, the resulting `ThinObj` is a temporary value (which can be quite confusing when you start borrowing things).

Perhaps surprisingly, this gives us an opportunity: we can't put casting on `Val` because if it's unboxed, we have to create a boxed version, which creates a temporary which we can't then borrow from. However, by putting `cast` on `ThinObj` we sidestep the problem entirely: we get a (relatively) nice way of casting (including generating a `VMError` which records what type we expected, and what we actually got) while avoiding introducing any new borrow checking problems.

### DIFF
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -33,7 +33,7 @@
 
 use std::{
     alloc::Layout,
-    any::Any,
+    any::{Any, TypeId},
     collections::HashMap,
     fmt::Debug,
     mem::{forget, size_of, transmute},
@@ -262,6 +262,17 @@ impl ThinObj {
     pub unsafe fn recover(o: &Obj) -> Gc<ThinObj> {
         let thinptr = (o as *const _ as *const u8).sub(size_of::<ThinObj>()) as *const ThinObj;
         Gc::recover(thinptr)
+    }
+
+    /// Cast this `ThinObj` to a concrete `Obj` instance.
+    pub fn cast<T: Obj + 'static>(&self) -> Result<&T, VMError> {
+        self.deref()
+            .as_any()
+            .downcast_ref()
+            .ok_or_else(|| VMError::TypeError {
+                expected: TypeId::of::<T>(),
+                got: self.deref().as_any().type_id(),
+            })
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/vm/objects.rs
```diff
@@ -263,6 +263,17 @@ impl ThinObj {
         let thinptr = (o as *const _ as *const u8).sub(size_of::<ThinObj>()) as *const ThinObj;
         Gc::recover(thinptr)
     }
+
+    /// Cast this `ThinObj` to a concrete `Obj` instance.
+    pub fn cast<T: Obj + 'static>(&self) -> Result<&T, VMError> {
```
jacob-hughes: Is `cast` the right terminology here? Intuitively something like `as_obj` makes more sense to me as casting implies the user will provide the destination type for conversion.

ltratt: A (too long but more accurate) name would be `cast_if_possible_or_return_vmerror_typeerror`. Dunno if that helps think of a better name or not ;)

src/lib/vm/objects.rs
```diff
@@ -263,6 +263,17 @@ impl ThinObj {
...
+        self.deref()
+            .as_any()
+            .downcast_ref()
```
jacob-hughes: So, this method should give us a fat pointer to some `T: Obj`?  It wouldn't hurt to add the type parameter `T` to `downcast_ref` for clarity.

ltratt: Yes, it gives us a fat pointer.

src/lib/vm/objects.rs
jacob-hughes: Why not `as_obj`?

jacob-hughes:
The code here looks fine, but I'm not quite clear on the motivation. The `Deref` trait for `ThinObj` constructs a fat pointer lazily for us which includes the vptr. Why might we want to cast to an `Obj`?

src/lib/vm/objects.rs
ltratt: Sorry, my answer was completely wrong -- I don't know what I was thinking :(

No, this doesn't give us back a fat pointer. It gives us a thin pointer to a concrete `struct` that `impl`s `Obj`.

jacob-hughes: Ah okay. This makes more sense now!

src/lib/vm/objects.rs
vext01: I'm OK with just `cast`

vext01:
This looks good as far as I understand the system, but admittedly I also don't really understand the motivation. It's not a criticism of you intention, but rather: I don't really know the system well enough to know what's going on here.

Perhaps tomorrow you can give me (or us) a rundown of how the system hang together?

ltratt:
I will try. I think we can merge this in the interim?

vext01:
It's OK for me, as long as Jake is convinced. He's the GC guy ;)

ltratt:
This isn't a GC PR.

vext01:
> I don't really know the system well enough to know what's going on here

I rest my case 🔨

ltratt:
OK, @jacob-hughes, it's down to you.

jacob-hughes:
LGTM, no edits so I'm guessing you don't need to squash? If so I'll merge

ltratt:
I can't squash 1 commit down any further :p

jacob-hughes:
You might have wanted to update commit message though ;)

ltratt:
That is true. But not in this case. Feel free to merge as-is!

PR #8, opened 2019-04-26
Add Gc::new.

ltratt:
This adds a simple `Gc::new` function which creates a new GC'd object of an arbitrary type (previously we only offered the indirect `GcBox::alloc` approach. This is likely to be useful in several places.

### DIFF
--- a/src/lib/vm/gc.rs
+++ b/src/lib/vm/gc.rs
@@ -80,14 +80,14 @@ impl<T: GcLayout> Gc<T> {
 
 impl<T: GcLayout> Drop for Gc<T> {
     fn drop(&mut self) {
-        let clones = unsafe { &mut *self.ptr }.clones;
+        let clones = unsafe { &*self.ptr }.clones;
         if clones == 1 {
             unsafe {
                 ptr::drop_in_place(self.ptr);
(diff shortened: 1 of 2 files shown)


jacob-hughes:
So I'm assuming that most users will use `Gc::new()`` to create new GC objects. If so, would it be worth no longer making `alloc_blank` public? That way we know that `from_raw` was immediately called by enforcing it through a common API. 

I'm guessing you've thought about this though, and have cases where `Gc::new()` won't always be appropriate.

ltratt:
> I'm guessing you've thought about this

It's safe to assume that I've often not thought about things :)

In this case, the two functions are doing something usefully different. `alloc_blank` allows you to reserve an arbitrarily sized chunk of memory and put whatever you want into it later. `new` reserves however much room `T` needs and copies a particular instance of `T` into it.

The cunning part about `alloc_blank` is that `GC` gives itself enough space to hide the reference count in as well as the space the user asked for -- all in one malloc, rather than the two you'd need to do if you used `new`.

jacob-hughes:
Ah I see, that makes sense! 

The code looks fine to me, I can't see anything in there which I would like changing. So if you're happy for me to merge I will.

ltratt:
Please go ahead!

PR #9, opened 2019-04-27
Search the class hierarchy for messages

ltratt:
This PR gets us to the point that we can write some mildly interesting programs. There are two major commits:

 * https://github.com/softdevteam/yksom/commit/146c8911244199cfd5da92e0962fdece21f3855a searches classes *and* super classes for a message `m` i.e. this implements inheritance.
 * https://github.com/softdevteam/yksom/commit/7fd95cc9b61ed174a2389538f894f7fc964f5dcb implements looking up of variables. Currently only the `self` variable is handled.

This is enough to allow us to write programs such as:

```
      run = ('woo' class asString println)
```

which, when run, prints out `String`. This might not look like much, but if you follow through what's going on, there's more than you might think (`'woo' class` returns the `String` class; `Strring asString` returns the string `String` and `println` prints it out).

The bad news is that I'm currently unhappy with the inefficient way we search the class hierarchy: it's not very efficient (we have to continually convert clas `Val`s to think objects, and cast them), and given how core this is likely to be to performance, we will need to do better. That's for a future PR.


vext01:
Looks good. A handful of comments.

vext01:
Looks like you successfully countered all of my comments. Are we ready to merge?

ltratt:
I think so!

PR #11, opened 2019-05-02
Faster casting

ltratt:
The main aim of this PR is to make casting from an `Obj` to a concrete struct faster (https://github.com/softdevteam/yksom/commit/654e044bbec3df77650e3f4fe334fa63c491e528). As a secondary aim, this also makes the boilerplate surrounding `Obj`s slightly less offensive (https://github.com/softdevteam/yksom/commit/ace48f549fe4033e13aad315aea16c3de69ff373), and then removes uses of the `Any` type entirely (https://github.com/softdevteam/yksom/commit/256c19c18861bc3545921798b09c2b89a2f84890). The individual commit messages -- particularly for the first commit -- hopefully explain things in sufficient detail.

### DIFF
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -266,13 +266,29 @@ impl ThinObj {
 
     /// Cast this `ThinObj` to a concrete `Obj` instance.
     pub fn cast<T: Obj + 'static>(&self) -> Result<&T, VMError> {
-        self.deref()
-            .as_any()
-            .downcast_ref()
-            .ok_or_else(|| VMError::TypeError {
+        // This is a cunning hack based on the fact that vtable pointers are a proxy for a type
+        // identifier. In other words, if two distinct objects have the same vtable pointer, they
+        // are instances of the same type; if their vtable pointers are different, they are
+        // instances of different types. We can thus use vtable pointers as a proxy for type
+        // equality.
+
+        // We need to get `T`'s vtable. We cheat, creating a dummy pointer that forces the compiler
+        // to create a trait object, from which we can then fish out the vtable pointer.
+        let t_vtable = {
+            let t: &dyn Obj = unsafe { &*(0 as *const T) };
+            unsafe { transmute::<&dyn Obj, (usize, usize)>(t) }.1
+        };
+
+        if t_vtable == self.vtable {
+            let self_ptr = self as *const Self as *const u8;
+            let obj_ptr = unsafe { self_ptr.add(size_of::<ThinObj>()) };
+            Ok(unsafe { &*(obj_ptr as *const T) })
+        } else {
+            Err(VMError::TypeError {
                 expected: TypeId::of::<T>(),
                 got: self.deref().as_any().type_id(),
             })
+        }
     }
 }
 
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -205,8 +205,8 @@ impl Drop for Val {
 
 /// The main SOM Object trait.
 pub trait Obj: Debug + GcLayout {
-    /// Return this object as an `Any` that can then be safely turned into a specific struct.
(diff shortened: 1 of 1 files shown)


vext01:
It's gross, but I like it. LGTM, but I think Jake has a query.

src/lib/vm/objects.rs
```diff
@@ -266,13 +266,29 @@ impl ThinObj {
 
     /// Cast this `ThinObj` to a concrete `Obj` instance.
     pub fn cast<T: Obj + 'static>(&self) -> Result<&T, VMError> {
-        self.deref()
-            .as_any()
-            .downcast_ref()
-            .ok_or_else(|| VMError::TypeError {
+        // This is a cunning hack based on the fact that vtable pointers are a proxy for a type
```
jacob-hughes: This is clever. Is this unique to SOM? For example, could I compare `vptrs` in Rust to determine type equality?

vext01: We were also wondering, a type which implements `Obj` would have a different table to one that implements `Obj` and (say) `Obj2`?

src/lib/vm/objects.rs
```diff
@@ -266,13 +266,29 @@ impl ThinObj {
...
+        // identifier. In other words, if two distinct objects have the same vtable pointer, they
+        // are instances of the same type; if their vtable pointers are different, they are
+        // instances of different types. We can thus use vtable pointers as a proxy for type
+        // equality.
+
+        // We need to get `T`'s vtable. We cheat, creating a dummy pointer that forces the compiler
+        // to create a trait object, from which we can then fish out the vtable pointer.
+        let t_vtable = {
+            let t: &dyn Obj = unsafe { &*(0 as *const T) };
+            unsafe { transmute::<&dyn Obj, (usize, usize)>(t) }.1
```
jacob-hughes: Ha, this is evil.

src/lib/vm/objects.rs
```diff
@@ -231,6 +231,14 @@ gclayout!(Inst);
 gclayout!(Int);
 gclayout!(String_);
 
+macro_rules! obj_boilerplate {
```
jacob-hughes: I'm not sure I understand why this can't be a default method in the `Obj` trait?

src/lib/vm/objects.rs
ltratt: > Is this unique to SOM?

No.

>For example, could I compare vptrs in Rust to determine type equality?

Yes. Although only if you're using trait objects and doing something like storing the vtable pointers manually.

> We were also wondering, a type which implements Obj would have a different table to one that implements Obj and (say) Obj2?

I'm not sure I understand the question?

src/lib/vm/objects.rs
ltratt: I haven't tried, but `Self` types in traits are a bit weird, so I assumed it wouldn't work. I may try later.

ltratt: Ah, no, I remember why: `type_id` on the trait returns the type ID of the trait, not the struct. The new commit's tests make that clear.

src/lib/vm/objects.rs
vext01: > I'm not sure I understand the question?

With a little thought, I answered it myself. Please ignore ;)

src/lib/vm/objects.rs
vext01: I was about to say: I've never managed to put a method which returns  `Any` in a trait. But then I managed it:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e2d7727193ad491aa8d091d3eee246c5

But I think discussion on `Any` is moot, since @ltratt removes all uses of it in a later commit.

ltratt:
I think this is ready to be squashed?

vext01:
Please squash.

ltratt:
Squashed.

PR #12, opened 2019-05-04
Get normal Rust references to objects when needed

ltratt:
The big problem with our previous `Val` scheme is that you had to call `Val::tobj()` in order to get something you could then turn into a `&T` normal Rust reference. Because `tobj` returns a temporary value (since it might have to box an unboxed value), this meant that you couldn't store references to a `Val`'s underlying object without jumping through hoops.

This PR fixes that, in two main stages. First, we introduce our own type ID scheme (https://github.com/softdevteam/yksom/commit/d42c95d51140a3a1d7d0eda52830e1cd7f12a896) instead of using `TypeId` (which uses the `'static` lifetime, which causes us various problems). Then we introduce a new function `Val::gcbox_cast` (https://github.com/softdevteam/yksom/commit/143e4f6b4020e1583b9030aafec2c682d879adfb) which allows one to cast a `Val` directly to a specific `&T` (where `T` is a struct implementing `Obj`). To be clear, this is generally dangerous, because `gcbox_cast` can't box things: if you call it on an unboxed int, for example, you'll get a `VMError`. But it's useful for things like storing direct Rust references to super classes. In other words, `gcbox_cast` should be used very sparingly.

### DIFF
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -86,7 +86,7 @@ impl Val {
     ///
     /// [In an ideal world, this would be a function on `Obj` itself, but that would mean that
     /// `Obj` couldn't be a trait object. Oh well.]
-    pub fn from_obj<T: Obj + 'static>(_: &VM, obj: T) -> Self {
+    pub fn from_obj<T: Obj>(_: &VM, obj: T) -> Self {
         debug_assert_eq!(ValKind::GCBOX as usize, 0);
         debug_assert_eq!(size_of::<*const GcBox<ThinObj>>(), size_of::<usize>());
         let ptr = ThinObj::new(obj).into_raw();
@@ -258,10 +258,10 @@ pub struct ThinObj {
 }
 
 impl ThinObj {
-    pub fn new<U>(v: U) -> Gc<ThinObj>
+    pub fn new<'a, U>(v: U) -> Gc<ThinObj>
     where
-        *const U: CoerceUnsized<*const Obj>,
-        U: Obj,
+        *const U: CoerceUnsized<*const (Obj + 'a)>,
+        U: Obj + 'a,
     {
         let (layout, uoff) = Layout::new::<ThinObj>().extend(Layout::new::<U>()).unwrap();
         debug_assert_eq!(uoff, size_of::<ThinObj>());
@@ -285,7 +285,7 @@ impl ThinObj {
     }
 
     /// Cast this `ThinObj` to a concrete `Obj` instance.
-    pub fn cast<T: Obj + StaticObjType + 'static>(&self) -> Result<&T, VMError> {
+    pub fn cast<T: Obj + StaticObjType>(&self) -> Result<&T, VMError> {
         // This is a cunning hack based on the fact that vtable pointers are a proxy for a type
         // identifier. In other words, if two distinct objects have the same vtable pointer, they
         // are instances of the same type; if their vtable pointers are different, they are
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -380,7 +380,7 @@ pub struct Class<'a> {
     pub name: Val,
     pub path: PathBuf,
     pub supercls: Option<&'a Class<'a>>,
-    pub methods: HashMap<String, Gc<Method>>,
+    pub methods: HashMap<String, Method>,
     pub instrs: Vec<Instr>,
     pub consts: Vec<Val>,
     pub sends: Vec<(String, usize)>,
@@ -428,7 +428,7 @@ impl<'a> Class<'a> {
                 name: cmeth.name.clone(),
                 body,
             };
-            methods.insert(cmeth.name, Gc::new(meth));
+            methods.insert(cmeth.name, meth);
         }
         let consts = ccls
             .consts
@@ -455,10 +455,10 @@ impl<'a> Class<'a> {
         Ok(self.name.clone())
     }
 
-    pub fn get_method(&self, vm: &VM, msg: &str) -> Result<(Val, Gc<Method>), VMError> {
+    pub fn get_method(&self, vm: &VM, msg: &str) -> Result<(Val, &Method), VMError> {
         self.methods
             .get(msg)
-            .map(|x| Ok((Val::recover(self), Gc::clone(x))))
+            .map(|x| Ok((Val::recover(self), x)))
             .unwrap_or_else(|| match self.supercls {
                 Some(scls) => scls.get_method(vm, msg),
                 None => Err(VMError::UnknownMethod(msg.to_owned())),
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -121,6 +121,26 @@ impl Val {
         }
     }
 
+    /// If this `Val` is a `GCBOX`, and that `GCBOX` is of type `T`, cast the `Val` to `&T`. If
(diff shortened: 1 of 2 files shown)


src/lib/vm/objects.rs
```diff
@@ -122,6 +121,26 @@ impl Val {
...
     }
 
+    /// If this `Val` is a `GCBOX`, and that `GCBOX` is of type `T`, cast the `Val` to `&T`. If
+    /// this `Val` is not a `GCBOX` or the `GCBOX` is not of type `T`, `VMError` will be returned.
+    /// Note that, in general, you should use `Val::tobj()` as that can box values as needed
+    /// whereas `gcbox_cast` cannot.
+    pub fn gcbox_cast<T: Obj + StaticObjType>(&self, _: &VM) -> Result<&T, VMError> {
+        match self.valkind() {
+            ValKind::GCBOX => {
+                debug_assert_eq!(ValKind::GCBOX as usize, 0);
```
jacob-hughes: Is this you being extra careful? Or are there places in the VM where it is quite possible that the tag bit might not match the enum variant?

ltratt: This is me being extra careful because the code doesn't do the necessary masking, since `GCBOX == 0`.

In retrospect, I could probably rely on the optimiser to remove the masking if `GCBOX == 0` which would remove the asserts. I would need to check that it actually does that -- I think it's something for a future PR.

src/lib/vm/objects.rs
```diff
@@ -568,6 +611,12 @@ impl Obj for String_ {
     }
 }
 
+impl StaticObjType for String_ {
```
jacob-hughes: Do you foresee additional methods being added to `StaticObjType` in the future? If not, I wonder if it would be a good idea to eliminate the boilerplate by implementing `static_objtype()` as part of a macro.

ltratt: Probably not, but since it would only save two lines per `impl` I must admit that I'm not *too* concerned about this -- the definition of the macro would probably mean that no code is saved overall, plus users need to understand whatever magic is underlying the macro.

ltratt:
> Grammar typo in commit

Oops!

ltratt:
Exactly!

jacob-hughes:
Nice, I like it.

ltratt:
> At the moment, I'm guessing because your Gc type is really just an Rc wrapper.

Yes, that's definitely part of it.

> Assuming, however, you swapped this with the language supported GC, this shouldn't have any effect on performance?

Yes, if we have a proper GC, we will no longer have the ref counting costs *but* we'll still have the casting costs. [Plus we'll have to have the GC explore that part of the graph. Whether that's expensive or not I have no idea yet.]

To be honest, the performance implications at the moment are pretty small either way, but I'm working on the marginal gains philosophy.

jacob-hughes:
> Yes, if we have a proper GC, we will no longer have the ref counting costs but we'll still have the casting costs.

Ah I see.

> [Plus we'll have to have the GC explore that part of the graph. Whether that's expensive or not I have no idea yet.]

I think the GC would have to explore that part of the graph anyway, assuming that there exists, in some way, a root to the containing `Class` or `Obj` struct.

ltratt:
> I think the GC would have to explore that part of the graph anyway, assuming that there exists, in some way, a root to the containing Class or Obj struct.

I think that's the case too, although I haven't fully convinced myself that there won't be cases where it won't be necessary. However, at least this way, we don't have to pay the (admittedly very) minor cost of converting the pointer back into something the GC can use.

jacob-hughes:
> However, at least this way, we don't have to pay the (admittedly very) minor cost of converting the pointer back into something the GC can use.

Agreed. 

LGTM (apart from minor typo). If you want to squash now I'll merge, but you might want to wait for Edd incase he had comments.

ltratt:
I'd best wait for Edd in case he spots anything else. Since this is one of the crucial parts of the VM, I'd like to make sure we get it right.

vext01:
So if I understand correctly, storing references to things under the current model would be annoying, since you'd have to keep the thin object around? Is that right?

vext01:
Typo in commit message:
> because the sole function it require prevents...

ltratt:
It's not so much that it's annoying (although it is, slightly), but more that the casting has a small but definite performance penalty that can't be statically optimised away.

ltratt:
Shall I rebase and fix that commit message?

src/lib/vm/objects.rs
```diff
@@ -121,6 +121,26 @@ impl Val {
         }
     }
 
+    /// If this `Val` is a `GCBOX`, and that `GCBOX` is of type `T`, cast the `Val` to `&T`. If
+    /// this `Val` is not a `GCBOX` or the `GCBOX` is not of type `T`, `VMError` will be returned.
+    /// Note that, in general, you should use `Val::tobj()` as that can box values as needed
+    /// whereas `gcbox_cast` cannot.
+    pub fn gcbox_cast<T: Obj + StaticObjType>(&self, _: &VM) -> Result<&T, VMError> {
```
vext01: Are you ever going to need a reference to other kinds of `val`s? Things which are not GCBOX?

I wonder if this could be generalised to an `as_ref` method?

I started thinking about this because the method name doesn't apply that we are making a ref...

src/lib/vm/objects.rs
```diff
@@ -429,13 +459,9 @@ impl Class {
...
             .get(msg)
             .map(|x| Ok((Val::recover(self), Gc::clone(x))))
-            .unwrap_or_else(|| {
-                let sc_tobj = self.supercls.tobj(vm)?;
-                if let Ok(scls) = sc_tobj.cast::<Class>() {
-                    scls.get_method(vm, msg)
-                } else {
-                    Err(VMError::UnknownMethod(msg.to_owned()))
-                }
+            .unwrap_or_else(|| match self.supercls {
```
vext01: OK, so now I see. This removes the need to cast for every method lookup.

Can the superclass change, or is it fixed?

src/lib/vm/objects.rs
```diff
@@ -763,4 +789,19 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn test_gcbox_cast() {
+        let vm = VM::new_no_bootstrap();
+        let v = String_::new(&vm, "s".to_owned());
+        assert!(v.gcbox_cast::<String_>(&vm).is_ok());
+        let v = Int::from_usize(&vm, 0).unwrap();
```
vext01: So why do we expect this to fail for an int? Because they don't get Gc Boxed?

vext01:
A couple of comments.

src/lib/vm/objects.rs
ltratt: As the doc comment hopefully (!) makes clear, this function can by definition only be used with `GCBOX`. It's a fundamental limitation.

src/lib/vm/objects.rs
ltratt: In SOM it's fixed.

src/lib/vm/objects.rs
vext01: OK

src/lib/vm/objects.rs
vext01: OK

src/lib/vm/objects.rs
vext01: OK

vext01:
I think we are good. Do your squashing and/or rebasing.

ltratt:
Commit message type fixed.

PR #15, opened 2019-05-16
Port to lang_tester.

ltratt:
This ports yksom to use `lang_tester`. Mostly -- happily! -- this involves deleting code.


ltratt:
Yep, this is ready as-is.

PR #16, opened 2019-05-21
Add support for instance vars

ltratt:
This adds support for instance variables. The main commit is https://github.com/softdevteam/yksom/commit/64769be20df3966b506c592bb451e0535839481c and the tests in there hopefully give a good idea about what's going on.

### DIFF
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -47,6 +47,13 @@ impl<'a> Compiler<'a> {
         let mut errs = vec![];
         let supercls = astcls.supername.map(|x| lexer.lexeme_str(&x).to_owned());
 
+        let mut inst_vars = HashMap::with_capacity(astcls.inst_vars.len());
+        for lexeme in &astcls.inst_vars {
+            let vars_len = inst_vars.len();
+            inst_vars.insert(lexer.lexeme_str(&lexeme), vars_len);
+        }
+        compiler.vars_stack.push(inst_vars);
+
         let mut methods = Vec::with_capacity(astcls.methods.len());
         for astmeth in &astcls.methods {
             match compiler.c_method(&astmeth) {
@@ -82,6 +89,7 @@ impl<'a> Compiler<'a> {
             name: lexer.lexeme_str(&astcls.name).to_owned(),
             path: compiler.path.to_path_buf(),
             supercls,
+            num_inst_vars: astcls.inst_vars.len(),
             methods,
             instrs: compiler.instrs,
             consts: compiler.consts.into_iter().map(|(k, _)| k).collect(),
@@ -162,10 +170,15 @@ impl<'a> Compiler<'a> {
                 self.vars_stack.push(vars);
                 let bytecode_off = self.instrs.len();
                 // We implicitly assume that the VM sets SELF_VAR to self.
-                for e in exprs {
+                for (i, e) in exprs.iter().enumerate() {
                     // We deliberately bomb out at the first error in a method on the basis that
                     // it's likely to lead to many repetitive errors.
                     self.c_expr(e)?;
+                    if i == exprs.len() - 1 {
+                        self.instrs.push(Instr::Return);
+                    } else {
+                        self.instrs.push(Instr::Pop);
+                    }
                 }
                 self.vars_stack.pop();
                 Ok(cobjects::MethodBody::User {
@@ -179,13 +192,13 @@ impl<'a> Compiler<'a> {
     fn c_expr(&mut self, expr: &ast::Expr) -> Result<(), Vec<(Lexeme<StorageT>, String)>> {
         match expr {
             ast::Expr::Assign { id, expr } => {
-                let id_str = self.lexer.lexeme_str(id);
-                let id_idx = *match self.vars_stack.last().unwrap().get(id_str) {
-                    Some(n) => n,
-                    None => return Err(vec![(*id, format!("Unknown variable '{}'", id_str))]),
-                };
+                let (depth, var_num) = self.find_var(&id)?;
                 self.c_expr(expr)?;
-                self.instrs.push(Instr::VarSet(id_idx));
+                debug_assert_eq!(self.vars_stack.len(), 2);
+                match depth {
+                    0 => self.instrs.push(Instr::InstVarSet(var_num)),
+                    _ => self.instrs.push(Instr::VarSet(var_num)),
+                }
                 Ok(())
             }
             ast::Expr::KeywordMsg { receiver, msglist } => {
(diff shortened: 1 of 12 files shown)


src/lib/compiler/ast_to_instrs.rs
```diff
@@ -162,10 +170,15 @@ impl<'a> Compiler<'a> {
                 self.vars_stack.push(vars);
                 let bytecode_off = self.instrs.len();
                 // We implicitly assume that the VM sets SELF_VAR to self.
-                for e in exprs {
+                for (i, e) in exprs.iter().enumerate() {
                     // We deliberately bomb out at the first error in a method on the basis that
                     // it's likely to lead to many repetitive errors.
                     self.c_expr(e)?;
+                    if i == exprs.len() - 1 {
```
vext01: Can you elaborate on this part? If it's the last expression, return else pop, but why?

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -221,15 +234,31 @@ impl<'a> Compiler<'a> {
...
+                match depth {
+                    0 => self.instrs.push(Instr::InstVarLookup(var_num)),
+                    _ => self.instrs.push(Instr::VarLookup(var_num)),
                 }
+                Ok(())
+            }
+        }
+    }
+
+    /// Find the variable `name` in the variable stack returning a tuple `Some((depth, var_num))`
```
vext01: So `vars_stack` is a stack of lexical scopes?

src/lib/vm/objects.rs
```diff
@@ -552,7 +559,24 @@ impl StaticObjType for Inst {
 
 impl Inst {
     pub fn new(vm: &VM, class: Val) -> Val {
-        Val::from_obj(vm, Inst { class })
```
vext01: I wonder if we should rename `Inst` to `Instance`.

It's caught me out a few times now. I always think `Instruction`.

src/lib/vm/objects.rs
```diff
@@ -524,6 +530,7 @@ impl StaticObjType for Method {
 #[derive(Debug)]
 pub struct Inst {
     class: Val,
+    inst_vars: UnsafeCell<Vec<Val>>,
```
vext01: I guess you were expecting me to ask about this line. Why the `UnsafeCell`? I'm not sure, and I think we need a comment?

src/lib/vm/vm.rs
```diff
@@ -185,11 +220,10 @@ impl VM {
                     pc += 1;
                 }
                 Instr::VarSet(n) => {
-                    let val = frame.stack_pop();
```
vext01: Is this an optimisation? It means you can use the top of the stack multiple times.

src/lib/vm/vm.rs
```diff
@@ -221,6 +255,10 @@ impl Frame {
         self.stack.push(v);
     }
 
+    fn stack_peek(&mut self) -> Val {
+        self.stack[self.stack.len() - 1].clone()
```
vext01: How about:
```
self.stack.last().unwrap().clone()
```

vext01: [I'll leave it to you if you decide to do something more user-friendly in the event of stack underflow]

vext01:
A handful of comments for ya.

src/lib/compiler/ast_to_instrs.rs
ltratt: At the end of each statement you have to clean up the stack. Think about code like `a = b = 0`.

src/lib/compiler/ast_to_instrs.rs
ltratt: Yes.

src/lib/vm/objects.rs
ltratt: Tomato, tomato ;) Anyway, even if we do, this isn't the right PR to do it.

src/lib/vm/objects.rs
ltratt: It's interior mutability. All objects in Rust are immutable, otherwise we can't write the VM sensibly. That means that doing things like updating a slot requires interior mutability. Since this part needs to be fast, `UnsafeCell` is our only realistic choice.

src/lib/vm/vm.rs
```diff
@@ -43,33 +43,54 @@ pub enum VMError {
...
+        // Nothing in this phase must store references to any classes earlier than it in the phase.
+        vm.str_cls = vm.init_builtin_class("String", false);
 
         vm
     }
 
-    /// Compile the file at `path`.
-    pub fn compile(&self, path: &Path) -> Val {
+    /// Compile the file at `path`. `inst_vars_allowed` should be set to `false` only for those
+    /// builtin classes which do not lead to run-time instances of `Inst`.
```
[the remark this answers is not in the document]
ltratt: See the comments on bootstrapping earlier in the file: does that make sense?

src/lib/vm/vm.rs
ltratt: It's to allow things like `a = b = 0`.

src/lib/vm/vm.rs
ltratt: Assuming the compiler is properly written, underflow should be impossible.

I suspect I should actually make this `get_unchecked` in release mode. Let me do that.

ltratt: The new commit does this (and a bit more).

src/lib/compiler/ast_to_instrs.rs
vext01: OK, is it worth a comment do you think?

src/lib/vm/objects.rs
vext01: > Tomato, tomato

Well in one sense yes, but that doesn't capture my gripe.

"Tom, Tom" would capture it ;)

The shortening of the name makes it ambiguous with other VM terms, no?

src/lib/vm/objects.rs
vext01: > Since this part needs to be fast, UnsafeCell is our only realistic choice.

OK, this is the missing detail. Shall we add a comment?

src/lib/vm/vm.rs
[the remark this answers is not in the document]
vext01: Ah, it's to do with that stuff. I see.

src/lib/vm/vm.rs
vext01: OK

src/lib/vm/vm.rs
vext01: Works for me.

src/lib/compiler/ast_to_instrs.rs
ltratt: I tried, but any comment I write just says what the code does. So I think it's best left alone.

src/lib/vm/objects.rs
ltratt: `Inst` is a common name in VMs for this concept, so it doesn't really trouble me. Anyway, this isn't something that should be/can be resolved in this PR.

src/lib/vm/objects.rs
ltratt: We're going to be using this pattern frequently soon enough, so no: it'll become really obvious what's going on I think.

src/lib/vm/objects.rs
vext01: I'm not sure I succeeded in communicating the problem, but I don't feel strongly enough to object further.

ltratt:
If you're OK with this PR, I'm fine with merging this as-is: the recent commit is nicely stand alone in its own way.

PR #19, opened 2019-06-08
Add support for booleans

ltratt:
This is the sort of PR I dislike reviewing, so apologies in advance, though each commit should be self-contained. It adds support for booleans, which turned out to be more involved than I expected.

https://github.com/softdevteam/yksom/commit/fb70abae4c58d9b776685ffea428fc471ee33608 adds the basic bool types. We then need to add support for blocks (i.e. anonymous functions) https://github.com/softdevteam/yksom/commit/713ac4b9b707dfdfa4efd2afd7e3cb7231a78b29 and closures (i.e. capturing a parent function's environment) https://github.com/softdevteam/yksom/commit/69a8bbba23ed4f14dfe187ee3ac063a4b308e9c8 and non-local returns (since SOM/Smalltalk use blocks/function calls for basic control flow) https://github.com/softdevteam/yksom/commit/ed522569224e19cd3068cd5e325458e409066c34. The other commits are mostly minor. By the end of this we can run some not completely useless programs such as https://github.com/softdevteam/yksom/pull/19/files#diff-d0f9b026092d7f61adca6daa28ff0cc9.


lib/SOM/False.som
```diff
@@ -0,0 +1,4 @@
+False = Boolean (
```
vext01: I'm not saying this is a bad idea, but what's the justification for "interning" (kind of) the true/false literals? I guess I expected `Boolean` to have a `val` field which could be (interp-level) true/false.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -234,11 +234,19 @@ impl<'a> Compiler<'a> {
...
-                    _ => self.instrs.push(Instr::VarLookup(var_num)),
+                match self.find_var(&lexeme) {
+                    Ok((depth, var_num)) => {
+                        debug_assert_eq!(self.vars_stack.len(), 2);
+                        match depth {
+                            0 => self.instrs.push(Instr::InstVarLookup(var_num)),
+                            _ => self.instrs.push(Instr::VarLookup(var_num)),
+                        }
+                    }
+                    Err(e) => match self.lexer.lexeme_str(&lexeme) {
```
vext01: You attempt to lookup a variable of the name "true" or "false". Does this mean that the tokens `true` and `false` are special, but not reserved?

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -160,16 +162,38 @@ impl<'a> Compiler<'a> {
                 exprs,
             } => {
                 let mut vars = HashMap::new();
-                const_assert_eq!(SELF_VAR, 0);
-                vars.insert("self", SELF_VAR);
-                for lexeme in vars_lexemes {
+                // The VM makes some strong assumptions about variables: that the first variable is
+                // "self" and that arguments 1..n+1 are the first n arguments to the method in
+                // reverse order.
```
vext01: So the reverse ordering is to do with copying to the stack?

lib/SOM/False.som
ltratt: This is similar to Python: `True` and `False` are singleton instances of `bool`. So they're not interned, but you want a single object to be `true` and `false` throughout the whole program.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -160,16 +162,38 @@ impl<'a> Compiler<'a> {
...
+                vars.insert("self", 0);
+
+                let mut process_var = |lexeme| {
                     let vars_len = vars.len();
-                    vars.insert(self.lexer.lexeme_str(&lexeme), vars_len);
+                    let var_str = self.lexer.lexeme_str(&lexeme);
+                    match vars.entry(var_str) {
```
vext01: I know you like to remove hashmaps wherever possible :)

Since most functions won't accept loads of arguments, I wonder if (later) a good strategy is to use a list of <name, index> pairs up until a threshold, where you revert to hashmaps?

Then again, the string comparisons themselves may be expensive enough to just use a hashmap from the outset...

src/lib/compiler/ast_to_instrs.rs
ltratt: Yes (SOM seems to behave the same too -- you can have a method name `true`, for example).

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -160,16 +162,38 @@ impl<'a> Compiler<'a> {
...
+                            format!("Variable '{}' shadows another of the same name", var_str),
+                        )]),
+                        hash_map::Entry::Vacant(e) => {
+                            e.insert(vars_len);
+                            Ok(vars_len)
+                        }
+                    }
+                };
+
+                for lexeme in args.iter().rev() {
```
vext01: Dumb question: what's the difference between these two loops?

src/lib/compiler/ast_to_instrs.rs
ltratt: Yes, in the long run, it will allow faster copying. At the moment, this isn't optimised as such though.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -160,16 +162,38 @@ impl<'a> Compiler<'a> {
...
+                    process_var(*lexeme)?;
+                }
+                for lexeme in vars_lexemes {
+                    process_var(*lexeme)?;
                 }
+
```
vext01: Do we need to check all variable slots are occupied to avoid undefined variables?

src/lib/compiler/ast_to_instrs.rs
ltratt: This is in the compiler, so there's no advantage to doing anything like that.

src/lib/compiler/ast_to_instrs.rs
ltratt: In SOM you have this sort of thing:

```
x: y = ( | z | ... )
```

so you have a parameter `y` (the first loop) and a local variable `z` (the second loop). They're both variables, but we have to collect them in different ways.

src/lib/compiler/ast_to_instrs.rs
ltratt: This is the compiler, so such checks don't really make sense.

lang_tests/block1.som
```diff
@@ -0,0 +1,9 @@
+"
+VM:
+  stdout:
+    Hello world
+"
+
+block1 = (
+    run = ( [ 'Hello world' ] value println )
```
vext01: Is the test runner looking for things called `run`?

ltratt: When SOM runs a program, it calls the class-specified-on-the-command-line's `run` method (think of it as `main`). So it's a general thing, not test runner specific.

src/lib/compiler/ast.rs
```diff
@@ -46,6 +46,11 @@ pub enum Expr {
         id: Lexeme<StorageT>,
         expr: Box<Expr>,
     },
+    Block {
+        params: Vec<Lexeme<StorageT>>,
```
vext01: Is a block's "parameters" the values it captures from the env?

ltratt: No, it literally can take in parameters.

lang_tests/block5.som
```diff
@@ -0,0 +1,19 @@
...
+VM:
+  stdout:
+    2
+"
+
+block5 = (
+    f = (
+       | x y |
+       x := '1'.
+       y := [ x ].
```
vext01: So x and y end up aliased? This wasn't what I expected.

vext01: Ah, the code block isn't evaluated yet.. My bad.

ltratt: No. `[ x ]` is a block (i.e. an anonymous function) which references `x`.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -311,17 +311,16 @@ impl<'a> Compiler<'a> {
...
+    /// "current" one that the variable is found.
     fn find_var(
         &self,
         lexeme: &Lexeme<StorageT>,
     ) -> Result<(usize, usize), Vec<(Lexeme<StorageT>, String)>> {
         let name = self.lexer.lexeme_str(lexeme);
         for (depth, vars) in self.vars_stack.iter().enumerate().rev() {
             if let Some(n) = vars.get(name) {
-                return Ok((depth, *n));
+                return Ok((self.vars_stack.len() - depth - 1, *n));
```
vext01: How comes we return the inverse of the depth now? Is that to do with the reverse order args?

src/lib/vm/vm.rs
```diff
@@ -326,17 +349,62 @@ impl Frame {
...
         }
     }
 
-    fn var_set(&mut self, var: usize, val: Val) {
-        self.vars[var as usize] = val;
+    fn var_set(&mut self, depth: usize, var: usize, val: Val) {
+        self.closure(depth).set_var(var, val);
+    }
+
+    fn closure(&self, mut depth: usize) -> Gc<Closure> {
```
vext01: Can we add comments to the funcs of this struct? Particularly `closure()`.

lang_tests/method2.som
```diff
@@ -0,0 +1,13 @@
...
+VM:
+  stdout:
+    instance of method2
+"
+
+method2 = (
+    f = ( 'irrelevant' )
+
+    run = (
+       self f asString println
```
vext01: So because we don't make any explicit return (`^`) we get `method2` itself?

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -299,12 +301,17 @@ impl<'a> Compiler<'a> {
...
             // it's likely to lead to many repetitive errors.
             self.c_expr(e)?;
-            if i == exprs.len() - 1 {
-                self.instrs.push(Instr::Return);
-            } else {
+            if i != exprs.len() - 1 {
                 self.instrs.push(Instr::Pop);
             }
         }
+        // Blocks return the value of the last statement, but methods return `self`.
```
vext01: Ugh..

src/lib/compiler/ast_to_instrs.rs
ltratt: It means we do the inverse once in the compiler (i.e. here) and not continuously at run-time. It's not a big deal either way admittedly.

src/lib/vm/vm.rs
ltratt: Point taken for `closure`. The others I'm less worried about at the moment, as they might be shuffled around, and are fairly obvious.

lang_tests/method2.som
ltratt: Yes. I was surprised by this too, but all these tests produce the same result in yksom as Java SOM!

src/lib/compiler/ast_to_instrs.rs
ltratt: Indeed...

vext01: Reading the next commit I think reveals why. To allow non-local returns, right?

Is there a way to return from a specific nesting level, or can you only return from the current frame (allowing it to implicitly return) or the outermost closure (via `^`)?

ltratt: Not exactly -- this is just SOM's (and presumably Smalltalk's) semantics. It's one of those things where it feels like a 70s language... I guess all languages have their warts.

Basically the block `['x']` evaluates to `'x'` but the block `[^'x']`, when evaluated, causes the *function* it is contained in to be returned.

lang_tests/block6.som
```diff
@@ -0,0 +1,16 @@
+"
```
vext01: The escaping block case is interesting. What do other implementations do? I suppose the frame from which we would return from changes to the "owner" of the closure?

ltratt: This is the same as Java SOM and I confirmed its intended semantics with Stefan. It wasn't entirely obvious to me what it was supposed to do before I asked him!

lib/SOM/False.som
vext01: Yes, but why. I suppose it saves memory, but are there other reasons too?

src/lib/compiler/ast_to_instrs.rs
vext01: Odd design decision, but if that's what we have to implement, so be it!

lib/SOM/False.som
ltratt: If you didn't do this, `True is True` might sometimes evaluate to `False`, which would be... weird!

src/lib/compiler/ast_to_instrs.rs
vext01: OK

src/lib/compiler/ast_to_instrs.rs
vext01: OK

src/lib/compiler/ast_to_instrs.rs
vext01: Again, odd. But out of our control, so OK.

ltratt: In fairness to SOM, this isn't odd: pretty much every language I can think of works like this.

src/lib/compiler/ast_to_instrs.rs
vext01: Can you elaborate on this?

Suppose Python allowed named arguments, but without defaults:
```
def f(a, b, c):
    ...

f(a=1, c=2)
```

`b` was not instantiated, so I'd expect some kind of error, be it runtime or compile time?

lang_tests/block1.som
vext01: Odd, but OK

src/lib/compiler/ast.rs
vext01: So what's the difference between a block and a closure in SOM? Is a closure a block?

lang_tests/block1.som
ltratt: It's no odder than calling `main` in most languages.

lang_tests/block5.som
vext01: Yes, understood, but I was confused by why it didn't yield `1`. It's because the block sees the update to `x` since it hasn't been evaluated just yet.

src/lib/compiler/ast.rs
ltratt: In my terminology (I'm not sure if SOM has terminology for this), evaluating a block creates a closure (sometimes called an "environment"). Importantly, the closure can outlast the evaluation of the block (if there are inner blocks etc.).

lang_tests/method2.som
vext01: Odd, but OK.

I'm spotting a trend here :)

lang_tests/block5.som
ltratt: > It's because the block sees the update to x since it hasn't been evaluated just yet.

I must admit, I don't know what that means!

src/lib/compiler/ast_to_instrs.rs
vext01: OK.

lang_tests/block6.som
vext01: OK.

lib/SOM/False.som
vext01: Good point, well made.

src/lib/compiler/ast_to_instrs.rs
vext01: I'm probably confused then. I'm still struggling with syntax a fair amount.

```
x: y = ( | z | ... )
```

The rhs is a block accepting one arg, the lhs defines a function `x` also accepting one arg?

so:
```
def x(y, z):
   ...
```

?

lang_tests/block1.som
vext01: I've never seen the ability to choose the starting function on the command line, but I guess it's flexible.

src/lib/compiler/ast.rs
vext01: OK.

lang_tests/block5.som
vext01: Sorry, I'm bad at explaining.

The first time round, I read `y := [x]` to boil down to `y := 1` which is wrong becuase (as you said above) it's more like `y := &x`. Via the reference, `y` observes the later `x := '2'`.

If you see what I mean ;)

src/lib/compiler/ast_to_instrs.rs
vext01: OK

src/lib/compiler/ast_to_instrs.rs
ltratt: Your best bet is to look at a Smalltalk primer, as this stuff is normal Smalltalk.

src/lib/compiler/ast_to_instrs.rs
ltratt: That's a run-time error in Python (and in SOM). This file is in the compiler.

lang_tests/block1.som
ltratt: Who said anything about choosing the starting function? It's like `java X` which calls `X.main`.

lang_tests/block5.som
ltratt: No, it's not like `y := &x`: it's like (in Python) `y=lambda: x` i.e.:

```
>>> x=2
>>> y=lambda:x
>>> x=3
>>> y()
3
```

src/lib/compiler/ast_to_instrs.rs
vext01: OK

src/lib/compiler/ast_to_instrs.rs
vext01: OK, there's the missing info. It crashes at runtime, so OK.

lang_tests/block1.som
vext01: > Who said anything about choosing the starting function?

We are mincing words now ;)

You can effectively choose which `run` function is the entry point.

Let's consider this resolved.

lang_tests/block5.som
vext01: Exactly.

This is what I meant by:
> It's because the block sees the update to x since it hasn't been evaluated just yet.

Resolved.

vext01:
I'm OK with this.

Do you want to do any squashing?

ltratt:
Yes, I will do a little bit. Bear with me.

ltratt:
Squashed.

vext01:
Oops. Sorry about the delay.

lib/SOM/False.som
smarr: Strictly speaking, there is no interning of booleans in SOM.

You can instantiate a fresh instance of `False` or `True`. And really, `true == True new` should never appear in code, because it's not idiomatic. `True new ifTrue: [...]` will work as expected.

src/lib/compiler/ast.rs
smarr: Smalltalk/SOM does not carefully distinguish the lexical occurrence from the dynamically created object.
You usually won't find the term "closure" as such in most Smalltalk VMs, though, you'll find [BlockClosure](http://www.mirandabanda.org/cogblog/2008/06/07/closures-part-i/) ([second example](https://www.gnu.org/software/smalltalk/manual-base/html_node/BlockClosure.html))

src/lib/compiler/ast_to_instrs.rs
smarr: Ehm, my memory is a bit hazy, but I don't think that's correct. In SOM, you should hit the [read global case](https://github.com/SOM-st/TruffleSOM/blob/master/src/trufflesom/compiler/Parser.java#L966) for `nil`, `false`, and `true` (and less known also `system`).

These globals are initialized as part of the bootstrapping: https://github.com/SOM-st/som-java/blob/master/src/som/vm/Universe.java#L398

So, this means these tokens are not special.

The only special tokens in the parser I recall at the moment are `self` and `super`.

src/lib/compiler/ast_to_instrs.rs
smarr: There is the desire to keep methods short (max. 7 lines), so, Smalltalk people never evolved the language beyond these simple mechanisms.

I am not sure, but Self may have opted to always return the value of the last expression.

The benefit of returning `self` implicitly is that you get fluent APIs everywhere. Something that is apparently hip in other languages now.

src/lib/compiler/ast_to_instrs.rs
ltratt: `true` and `false` aren't special tokens in yksom, they're just variables. However, they are recognised as special variables.

lib/SOM/False.som
ltratt: I guess it largely depends on whether you can do identity equality checks: if you can, then instantiating new `true`s will end in disaster :) yksom allows you to instantiate the classes as many times as you want, and it should do the same as normal SOM (if I haven't made any mistakes!).

src/lib/compiler/ast.rs
ltratt: > Smalltalk/SOM does not carefully distinguish the lexical occurrence from the dynamically created object.

Aha, that would probably explain why I couldn't work out what the terminology was :)

lib/SOM/False.som
smarr: It's a Smalltalk, of course you can ;)
You can shoot yourself in the foot, the leg, or head. As much as you desire.

If you want even more power, consider implementing Smalltalk's `become:`.
And do `true become: false`. Always fun :)

Smalltalk is not designed to be a safe language. It aspires to be a systems language.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -234,29 +251,96 @@ impl<'a> Compiler<'a> {
...
+                        if depth == self.vars_stack.len() - 1 {
+                            self.instrs.push(Instr::InstVarLookup(var_num));
+                        } else {
+                            self.instrs.push(Instr::VarLookup(depth, var_num));
+                        }
+                    }
+                    Err(e) => match self.lexer.lexeme_str(&lexeme) {
+                        "nil" => self.instrs.push(Instr::Builtin(Builtin::Nil)),
+                        "false" => self.instrs.push(Instr::Builtin(Builtin::False)),
+                        "true" => self.instrs.push(Instr::Builtin(Builtin::True)),
```
smarr: The above three lines look wrong.

`nil`, `false`, `true` should compile to `pushGlobal(Symbol)`.

One can do `system global: #true put: false`. (so, you don't even need `#become:`, but `#become:` is more fun)

See https://github.com/smarr/SOM/blob/master/Smalltalk/System.som#L30

ltratt: This is pretty much what these lines are doing (this is the compiler: you can see the runtime part at https://github.com/softdevteam/yksom/pull/19/files#diff-4d6c08299827621fb4ed1166405a467dR242). Admittedly at the moment you can't update the builtins/globals. That's something for the future and, hopefully, at that point the semantics should align with SOM.

smarr: hm, I would expect something like the following here:

`str => self.instrs.pushGlobal(convertToSymbol(str))`.

How do you deal with classnames?
They go through this mechanism as well.

An expression such as `True new` would do:

```
pushGlobal(#True)
sendMessage(#new)
```

ltratt: At the moment there isn't a global lookup mechanism for classes such as `True` (at the moment you get an `Unknown variable 'True'` error). When there is, `True new` will, I'm fairly sure, become equivalent to the code you've mentioned without me doing anything clever.

smarr: Ok. Just to explain how SOM works, assume the following code:

```Smalltalk
MyClass = (
  myMethod = (
    FooBar println.
  )
)
```

When you would execute `MyClass new myMethod`, this would trigger a `#unknownGlobal:` on `self`, i.e., the `MyClass` instance.

The implementation in [Object](https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L81) then asks the system to check whether there is a class with the corresponding name on the classpath, and tries to load it.

And, to extend the example, in a spec-y way:

```Smalltalk
MyClass = (
  myMethod = (
    FooBar println.
  )

  unknownGlobal: name (
    ^ 'baz'
  )
)
```

With this method added, you would get `MyClass new myMethod` to output `baz\n`.

ltratt: Aha, interesting. As you have no doubt realised, I haven't got to that part of the semantics yet :)

PR #20, opened 2019-06-08
Don't use cargo run in lang tests.

ltratt:
`cargo run` has a noticeable overhead of about 0.1s vs. running a binary directly. Normally this isn't a problem worth bothering about, but since we're starting to grow an ever larger number of language tests, this starts to add up. On my main development branch (which has 22 language tests) on a 4 core machine, this cuts the testing time down from 3s to 2s (if I force only 1 core to be used, it cuts the time down from 10s to 7s, so this patch is more advantageous the fewer cores you have). That might not sound like much, but we're only going to grow more and more language tests over time...

### DIFF
--- a/lang_tests/run.rs
+++ b/lang_tests/run.rs
@@ -7,7 +7,7 @@
 // at your option. This file may not be copied, modified, or distributed except according to those
 // terms.
 
-use std::process::Command;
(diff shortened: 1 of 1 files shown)


vext01:
This is fine.

PR #21, opened 2019-06-10
Not using "dyn" for trait objects is now deprecated.

ltratt:
Fixing these (thanks to `cargo fix`) removes a large number of warnings.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -12,7 +12,7 @@ use lrlex::LexerBuilder;
 use lrpar::CTParserBuilder;
 use rerun_except::rerun_except;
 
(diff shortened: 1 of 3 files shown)


ltratt:
Wait just a second. There's one more I've missed.

ltratt:
Sorry, `cargo fix` didn't get the one in the newest commit. If you're OK with this, it'll need squashing.

vext01:
Not much to object to here! I've been through the same with ykpack recently.

Squash away.

ltratt:
Squashed.

PR #22, opened 2019-06-30
Val's cannot contain references.

ltratt:
We previously, unintentionally, undermined Rust's type system: you could have an object with lifetime 'a, move it into a `Val`, and the lifetime disappeared. This commit minimally fixes the only instance of this (in `Class`) and prevents anyone doing this again by having `Val::from_obj` force a `'static` lifetime on anything from which a `Val` is created.

### DIFF
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -87,7 +87,7 @@ impl Val {
     ///
     /// [In an ideal world, this would be a function on `Obj` itself, but that would mean that
     /// `Obj` couldn't be a trait object. Oh well.]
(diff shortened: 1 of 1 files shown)


vext01:
This seems smart to me.

Do you have any idea if requiring only `'static` references will be a pain for the VM author? I guess we won't know until later?

ltratt:
It is a pain, but I've currently had three goes at relaxing that constraint and failed. I have an idea for a fourth, but each attempt is at least half a day's effort...

vext01:
> It is a pain, but I've currently had three goes at relaxing that constraint and failed. I have an idea for a fourth, but each attempt is at least half a day's effort...

Sure, no worries. I was just wondering.

PR #23, opened 2019-07-24
Use abgc

ltratt:
This PR is a just-a-bit-more-than-minimal move to using the abgc crate. The first commit minimally adjusts yksom; the second commit then uses the `GcLayout` derive feature to (mostly) simplify the code. However, there is a catch: since `GcLayout` is now in an external crate we can't do things like `Gc<X>` if `X` is defined outside this crate. Instead we have to define a new `struct S(X)` and then say `Gc<S>` which is annoying but this is a normal Rust restriction, so we can't do anything about it.


vext01:
I've done a shallow review. LGTM.

Shall we keep the commits separate?

vext01:
Discussed on MM, we will keep them separate.

PR #24, opened 2019-07-25
Move from enum-primitive-derive to num_enum.

ltratt:
enum-primitive-derive isn't maintained and doesn't have a very nice way of converting a number into an enum. num_enum is a little better in that regard and is actively maintained, so we might even be able to drop the try_into at some point.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,10 +32,10 @@ regex = "1.1"
 abgc = { git = "https://github.com/softdevteam/abgc" }
 abgc_derive = { git = "https://github.com/softdevteam/abgc" }
 cfgrammar = "0.4"
-enum-primitive-derive = "0.1"
 getopts = "0.2"
 indexmap = "1.0"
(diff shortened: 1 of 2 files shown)


ltratt:
That new commit removes a small bottle-neck from the fast path. This will need squashing before merging.

vext01:
Maybe I've misunderstood, but if you are not going to check the conversion, why don't you use the built-in casting using `as`, described here:
https://stackoverflow.com/questions/33354058/how-do-i-get-the-integer-value-of-an-enum

Or put differently, is there a difference between `as` and `from_unchecked`? The latter boils down to an unsafe transmute, not sure about the former.

ltratt:
The new commit does the opposite direction: it turns an integer into the enum without doing any checks.

vext01:
I see.

I'm OK with this, so please squash.

ltratt:
Squashed.

PR #25, opened 2019-08-06
Port to the natrob library.

ltratt:
The top-level `downcast` function isn't hugely elegant, but it is simple: we might be able to find a nicer way of doing this in future.

### DIFF
--- a/src/lib/vm/vm.rs
+++ b/src/lib/vm/vm.rs
@@ -15,7 +15,7 @@ use std::{
 
 use abgc::{Gc, GcLayout};
 
-use super::objects::{Block, Class, Inst, MethodBody, ObjType, String_, Val};
+use super::objects::{downcast, Block, Class, Inst, MethodBody, ObjType, String_, Val};
 use crate::compiler::{
     compile,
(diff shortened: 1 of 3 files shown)


src/lib/vm/vm.rs
```diff
@@ -251,12 +254,12 @@ impl VM {
                     pc += 1;
                 }
                 Instr::InstVarLookup(n) => {
-                    let inst: &Inst = rcv.gcbox_cast(self).unwrap();
+                    let inst: &Inst = rcv.gcbox_downcast(self).unwrap();
```
vext01: In a prior PR, we removed a `GcBox` type. Is this function appropriately named? I guess natrob has its own notion of a gcbox?

vext01:
One small comment.

src/lib/vm/vm.rs
ltratt: yksom has a type `GCBOX` of its own, so this function is appropriately named in this context.

vext01: OK

PR #26, opened 2019-08-07
Simplify casting.

ltratt:
This is a deceptively simple PR and I can't believe it didn't occur to me earlier. In essence, we add a new trait `NotUnboxable` which if implemented for a SOM object means "this struct can never be unboxed". In our case, we implement this trait for everything except `Int`s. If you want to downcast a `Val` to a given `NotUnboxable` object type you can then change:

```
    let str_tobj = rcv.tobj(self)?;
    let str_: &String_ = downcast(&str_tobj)?;
```

into:

```
    let str_: &String_ = rcv.gcbox_downcast(vm)?;
```

Not only is this much simpler to write and read, but a) we can now statically detect when you're trying to do an illegal `gcbox_downcast` (e.g. you get a compile-time error if you try to downcast to an `Int`) b) we remove a check and branch from the performance critical `gcbox_downcas` function. So this seems like a win all round!

### DIFF
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -69,6 +69,9 @@ enum ValKind {
     // All of the values here must fit inside TAG_BITSIZE bits and be safely convert to usize
     // using "as".
     GCBOX = 0b000,
+    // Anything which can be stored unboxed *must* not have the `NotUnboxable` trait implemented
+    // for them. In other words, if an existing type is added to the list of unboxable things, you
+    // need to check whether it implemented `NotUnboxable` and, if so, remove that implementation.
     INT = 0b001,
 }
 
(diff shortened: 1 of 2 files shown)


vext01:
After getting over the initial upside-down thinking phase, this seems sensible to me. The added static checking is definitely good!

ltratt:
Originally I had hoped to use negative trait bounds (so every `Obj` would be boxable unless you explicitly opted out) but those simply aren't implemented fully enough.

I should have given a tip of the hat to @jacob-hughes here, because the `NotTrait` idea is vaguely based on ideas he had/has in his GC trait. I agree that it seems a bit upside-down at first!

vext01:
Unless you want input from @jacob-hughes(?), I'll merge this.

ltratt:
No input needed from him -- just me saying thanks!

jacob-hughes:
LGTM anyway :)

PR #28, opened 2019-08-30
Break out the Val struct (and friends)

ltratt:
This is a non-functional change, but it hopefully makes the code easier to read: it moves the `Val` struct (and some associated bits) into its own module. `objects.rs` then becomes more clearly "SOM objects" and `val.rs` "cunning optimisations to squeeze everything into a machine word". I've done this over several commits to make it easier to review.

### DIFF
--- a/src/lib/vm/mod.rs
+++ b/src/lib/vm/mod.rs
@@ -8,6 +8,7 @@
 // terms.
 
 pub mod objects;
+pub mod val;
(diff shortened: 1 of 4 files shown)


vext01:
LGTM

vext01:
I take it we we leave the commits separate?

ltratt:
Leave as-is please.

PR #29, opened 2019-09-03
Remove duplicate comment.

ltratt:
This comment is also present (as it should be) in objects.rs, which makes a great deal more sense.


vext01:
Has CI stalled again?

ltratt:
Nope, I just hadn't quite got the right settings in GitHub. It should be correct now!

PR #30, opened 2019-09-03
Use a custom, compact representation of Result<Val, Box<VMError>>.

ltratt:
The problem we had before is that `VMError` is, naturally, a fairly large enum: on a 64 bit machine it is currently 32 bytes. Before this commit, we have lots of functions with a return type of `Result<Val, VMError>` -- which is 40 bytes big! This is going to cause things to continually spill to the stack, which is not good for performance.

This commit does two things (in one go, because it made more sense that way): it changes `Result<Val, VMError>` into `Result<Val, Box<VMError>>` and then makes a new type `ValResult` which is a custom representation of that type. `ValResult` is a machine word big, so in essence this commit turns the 40 bytes of `Result<Val, VMError>` into the 8 bytes of `ValResult`.

The underlying technique is pretty simple: we have spare tag bits in `Val`, so we can use one of those to differentiate between a `Val` and a `ValResult`. This means that we get compile-time checking that we're not using a `Val` where we wanted a `ValResult`, but the two both share the same underlying representation (a machine word).

The main commit here is https://github.com/softdevteam/yksom/commit/4e7e230febc92c66ff8131c7a1a69bb5d2c2b70c; https://github.com/softdevteam/yksom/commit/f9ad08635444745374cb54256b53a5d5c368a513 breaks out some unsafe stuff so that it's more clearly visible for review; and https://github.com/softdevteam/yksom/commit/5da4f75fa66c6634de982c85b0f9312033511ce5 seems just about related enough to be worth including here (but feel free to tell me it should be in a separate PR).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,37 @@ lrpar = "0.4"
 num_enum = "0.3"
 num-traits = "0.2"
 natrob = { git="https://github.com/softdevteam/natrob", features=["abgc"] }
+
+# The main reason for customising the profile.* sections is to force unwind=abort.
+[profile.release]
+opt-level = 3
+debug = false
+rpath = false
+lto = true
+debug-assertions = false
+codegen-units = 1
+panic = 'unwind'
+incremental = false
+overflow-checks = false
+
+[profile.dev]
+opt-level = 0
+debug = true
+rpath = false
+lto = false
+debug-assertions = true
+codegen-units = 16
+panic = 'abort'
+incremental = true
+overflow-checks = true
+
+[profile.test]
+opt-level = 0
+debug = 2
+rpath = false
+lto = false
+debug-assertions = true
+codegen-units = 16
+# Currently, the panic setting is ignored for the 'test' profile.
+incremental = true
+overflow-checks = true
--- a/src/lib/vm/mod.rs
+++ b/src/lib/vm/mod.rs
@@ -13,7 +13,7 @@ macro_rules! vtry {
     ($elem:expr) => {{
         let o = $elem;
         if o.is_val() {
-            o.unwrap()
+            unsafe { o.unwrap_unsafe() }
         } else {
             return o;
         }
--- a/src/lib/vm/mod.rs
+++ b/src/lib/vm/mod.rs
@@ -7,6 +7,32 @@
 // at your option. This file may not be copied, modified, or distributed except according to those
 // terms.
 
+/// If a `ValResult` represents an error, percolate it up the call stack.
+#[macro_export]
+macro_rules! vtry {
+    ($elem:expr) => {{
+        let o = $elem;
+        if o.is_val() {
+            o.unwrap()
+        } else {
+            return o;
+        }
+    }};
+}
+
+/// If a `Result<T, Box<VMError>>` represents an error, then convert the `Box<VMError>` into a
+/// `ValResult` and percolate that up the call stack.
+#[macro_export]
+macro_rules! rtry {
+    ($elem:expr) => {{
+        let e = $elem;
+        match e {
+            Ok(o) => o,
+            Err(e) => return ValResult::from_boxvmerror(e),
+        }
+    }};
+}
+
 pub mod objects;
 pub mod val;
 pub mod vm;
--- a/src/lib/vm/objects.rs
+++ b/src/lib/vm/objects.rs
@@ -43,7 +43,7 @@ use crate::{
         instrs::{Instr, Primitive},
     },
     vm::{
-        val::{NotUnboxable, Val},
+        val::{NotUnboxable, Val, ValResult},
         vm::{Closure, VMError, VM},
     },
 };
@@ -68,9 +68,9 @@ pub trait Obj: Debug + abgc::GcLayout {
     /// Return the `ObjType` of this object.
     fn dyn_objtype(&self) -> ObjType;
     /// If possible, return this `Obj` as an `isize`.
-    fn as_isize(&self) -> Result<isize, VMError>;
+    fn as_isize(&self) -> Result<isize, Box<VMError>>;
     /// If possible, return this `Obj` as an `usize`.
-    fn as_usize(&self) -> Result<usize, VMError>;
+    fn as_usize(&self) -> Result<usize, Box<VMError>>;
     /// What class is this object an instance of?
     fn get_class(&self, vm: &VM) -> Val;
 }
@@ -92,12 +92,12 @@ impl Obj for Block {
         ObjType::Block
     }
 
-    fn as_isize(&self) -> Result<isize, VMError> {
-        Err(VMError::CantRepresentAsUsize)
+    fn as_isize(&self) -> Result<isize, Box<VMError>> {
+        Err(Box::new(VMError::CantRepresentAsUsize))
     }
 
-    fn as_usize(&self) -> Result<usize, VMError> {
-        Err(VMError::CantRepresentAsUsize)
+    fn as_usize(&self) -> Result<usize, Box<VMError>> {
+        Err(Box::new(VMError::CantRepresentAsUsize))
     }
 
     fn get_class(&self, vm: &VM) -> Val {
@@ -157,12 +157,12 @@ impl Obj for Class {
         ObjType::Class
     }
 
-    fn as_isize(&self) -> Result<isize, VMError> {
-        Err(VMError::CantRepresentAsUsize)
+    fn as_isize(&self) -> Result<isize, Box<VMError>> {
+        Err(Box::new(VMError::CantRepresentAsUsize))
     }
 
-    fn as_usize(&self) -> Result<usize, VMError> {
-        Err(VMError::CantRepresentAsUsize)
+    fn as_usize(&self) -> Result<usize, Box<VMError>> {
+        Err(Box::new(VMError::CantRepresentAsUsize))
     }
 
     fn get_class(&self, vm: &VM) -> Val {
@@ -179,7 +179,7 @@ impl StaticObjType for Class {
 }
 
 impl Class {
-    pub fn from_ccls(vm: &VM, ccls: cobjects::Class) -> Result<Val, VMError> {
+    pub fn from_ccls(vm: &VM, ccls: cobjects::Class) -> ValResult {
         let supercls = match ccls.supercls {
             Some(ref x) => match x.as_str() {
                 "Boolean" => Some(vm.bool_cls.clone()),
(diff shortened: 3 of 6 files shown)


vext01:
> This is going to cause things to continually spill to the stack, which is not good for performance.

Have you looked at some asm to verify this? It wouldn't surprise me if it did some magic (remember the optimisations Gabi found for large return types?).

ltratt:
Indeed, this PR is based on Gabi and Jake's observations that large return values (and I'm pretty sure that a 40 byte `Result` struct comes under this heading) are spilled to the stack (in the sense that space is reserved for them on the caller stack). I admit that I haven't checked the asm, because I don't know how to do that well in a case like this.

vext01:
> (in the sense that space is reserved for them on the caller stack)

But that's super cheap! It'll be part of the frame activation which is a `SUB %rsp, 0x...`.

That's much cheaper than a heap allocation!

ltratt:
Yes, but the heap allocation case is the rare case here, so this PR is optimising the common case which is that you're percolating a `Val` up the call stack. Basically the common case now involves returning a machine word, which should (in most cases, one hopes) be a return in a register, which is going to be much quicker than writing 40 bytes to memory, returning, and then reading some of those 40 bytes (even if those bytes are in L1). Because we currently have no practical way to benchmark, this is a little theoretical, though, I admit.

There is one exception to this, which I completely forgot to mention in the commit message: at the moment SOM closures (not functions) percolate values up the call stack by abusing `VMError` so they do now turn into a heap allocation. I believe I have a sensible way of completely avoiding that abuse, but that's for a future PR. If we do want to merge this PR, I should update the commit message to make this limitation clear.

vext01:
> the heap allocation case is the rare case here

This was going to be my next question. In light of this, I'm probably (not actually looked at the code yet) OK with this.

Out of curiosity, I made an example and read the (`-C opt-level=2`) code:

image

The return value is written indirectly via `r14`, which is a pointer to the parent's stack.

> Basically the common case now involves returning a machine word, which should be a return in a register, which is going to be much quicker than writing 40 bytes to memory.

You are close. Only the size of the `Ok(...)` is written to the parent stack. Nonetheless, as you say, it's still going to be faster to use a register. 

> which should (in most cases, one hopes) be a return in a register,

Well, let's find out. Same example using a box:

image

It writes the result (again) to the parent stack, but also returns a pointer to it 😆 

Despite that weirdness (let's assume it's a bug), I think your strategy is more likely to be optimisable in general.

ltratt:
Your example isn't quite the same though? I've turned `Result<Val, VMError>` (40 bytes) into `ValResult` (8 bytes).

vext01:
> Your example isn't quite the same though

Gah, yeah, I've just realised.

But wait, I'm slightly confused:

> The problem we had before is that VMError is, naturally, a fairly large enum: on a 64 bit machine it is currently 32 bytes. Before this commit, we have lots of functions with a return type of Result<Val, VMError>. Before this commit, we have lots of functions with a return type of Result<Val, VMError> -- which is 40 bytes big!

`VMError` is 32 bytes and `Val` is 8, then why is `Result<Val, VMError>` 40 bytes? Isn't the `Result` enum going to be the upper bound of its constituents? In which case `Result<Val, Box<VMError>>` suffices to fit in a register?

I must have messed up somewhere.

ltratt:
`Result<T, E>` needs to have some sort of tag to tell if it's storing a `T` or an `E`. If you're lucky, `T` and `E` have a spare bit that you can use (e.g. if both are pointers). If not, you need a whole extra machine word to store the tag. Since `VMError` has no spare bits, you end up with 40 bytes overall.

vext01:
I see.

ltratt:
If you are able to look at the asm before/after this PR, I would find that very interesting! I'm afraid that entirely exceeds my asm-finding-and-looking abilities!

src/lib/vm/mod.rs
```diff
@@ -7,6 +7,32 @@
 // at your option. This file may not be copied, modified, or distributed except according to those
 // terms.
 
+/// If a `ValResult` represents an error, percolate it up the call stack.
+#[macro_export]
+macro_rules! vrtn {
```
vext01: What's the `tn` in `vrtn`? Value result something something?

vext01: Or is it: something ReTurN?

This needs a better name basically ;)

ltratt: `ValReturn` (and `rrtn` is `ResultReturn`). Not that I think it's *too* important -- they just need short names, because they're used frequently and because, unfortunately, the `Try` trait isn't appropriate for our case (so we can't overload `?`).

src/lib/vm/mod.rs
```diff
@@ -7,6 +7,32 @@
...
+    ($elem:expr) => {{
+        let o = $elem;
```
vext01: Is this `let` binding needed?

src/lib/vm/mod.rs
vext01: We must be able to find something a little less cryptic. `val_ret!`?

Cargo.toml
```diff
@@ -40,3 +40,36 @@ lrpar = "0.4"
 num_enum = "0.3"
 num-traits = "0.2"
 natrob = { git="https://github.com/softdevteam/natrob", features=["abgc"] }
+
+[profile.release]
+opt-level = 3
```
vext01: Are you re-stating a load of defaults here, or are they values explicitly chosen to interact with the abort strategy somehow? Can defaults be omitted?

src/lib/vm/mod.rs
ltratt: It's to prevent the expression being evaluated twice (an old macro problem that you can also get in C).

Cargo.toml
ltratt: I'm unsure -- the documentation is unclear. Restating them seemed the safer thing...

src/lib/vm/mod.rs
ltratt: Too long IMHO.

vext01:
The change looks OK, but it adds a lot of complexity.

I wonder if we are going to end up with a LuaJIT-style VM with lot's of bit-packing and subtle implementation details. Correct me if I'm wrong!

src/lib/vm/mod.rs
vext01: We are going to have to find something less cryptic than `vrtn` and `rrtn`. I'm struggling to even type them into this box.

ltratt:
You're not wrong: one of the lessons I have taken from LuaJIT is that, particularly in the interpreter, you win big from cunning bit packing. What I'm trying to do in `yksom` is to isolate the nasty stuff to a couple of places (`Val` and `ValResult`) rather than spread itaround. Is it perfect? Certainly not. But I don't think it's *that* bad in (say) `objects.rs`.

src/lib/vm/mod.rs
vext01: Hah! Well I never.

src/lib/vm/mod.rs
ltratt: I'm open to suggestions, but not if they're very long. These macros will end up being called everywhere. Maybe `vtry` and `rtry`, which is a better mirror of `try` (which, in spirit, is what they resemble)?

Cargo.toml
vext01: OK, can we add a comment saying why we are adding the custom profiles?

vext01:
> Is it perfect? Certainly not. But I don't think it's that bad in (say) objects.rs.

OK. I just wanted to take a quick reality check to see if we were on the same page. Let's continue and see where it lands us.

Cargo.toml
ltratt: Fair enough -- see https://github.com/softdevteam/yksom/pull/30/commits/2cc2613eeee353871118874f07922a5aa0989d83

ltratt:
> Let's continue and see where it lands us.

I think that's the right strategy. I don't know for sure if this is the right structure. Perhaps I'm barking up the wrong tree but we probably won't know until we've got further down the line. But we can keep adjusting as we learn more.

vext01:
> If you are able to look at the asm before/after this PR, I would find that very interesting!

I can have a go at this if you want. But probably after lunch. Or if you wanted to try, the flag you want is `--emit asm`.

ltratt:
Honestly, I'd drown in the asm if I tried :/ I'm not really an x86 person...

src/lib/vm/mod.rs
vext01: At least with `vtry` and `rtry` I can discern the word boundaries :)

ltratt: https://github.com/softdevteam/yksom/pull/30/commits/ce7ca3bf33a7e38492fe25134114046d24988863 does this. I prefer the new names, so that's a win!

vext01:
I think that's all comments covered? Time to squash?

ltratt:
Squashed.

smarr:
> at the moment SOM closures (not functions) percolate values up the call stack by abusing VMError so they do now turn into a heap allocation.

I am stumbling over this. I might be misreading, ignoring relevant context, but I feel obliged to say: SOM closure objects (instances of the Block class) are supposed to be heap allocated.
Though, since I don't know what exactly the "percolate values up the call stack" is about, it may be entirely irrelevant.

vext01:
Waiting for Stefan's comment to be addressed prior to merge.

ltratt:
The closure is heap allocated, but when you do a return in the closure, you don't *need* to do an intermediate heap allocation in order to get the return value to the right place (if that makes sense).

smarr:
Sorry, I probably shouldn't discuss this here.

My impression was yksom is a SOM bytecode interpreter, though, it does not seem to be the case.
While there seems to be an AST to instruction transformation, the instruction set seems incompatible:
https://github.com/softdevteam/yksom/blob/master/src/lib/compiler/instrs.rs#L12

SOM has 16 bytecodes: https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Bytecodes.java#L34

One of which is a`NON_LOCAL_RETURN`. It's almost the same as a return in the relevant bits, and boils down to:

```java
SAbstractObject result = getFrame().pop();
// ... handle non-local stuff ...
popFrameAndPushResult(result);
```

As such, there is no heap allocation, instead there is a modification of a heap object (the frame, some Smalltalks call it context) representing the activation record.

ltratt:
I'm fine discussing this wherever! There's been no intention, currently, to make things bytecode compatible with normal SOM because -- as far as I know -- the bytecodes aren't visible to the SOM user?

smarr:
Bytecodes are not visible, no. So, you're indeed free to change them. SOM++ for instance adds some for branches.

Though, it's a little sad that you don't want to stick to the bytecode set.
That will add another variable into any comparisons 🙁

ltratt:
Understood. It's something we may come back to at some point perhaps.

PR #32, opened 2019-09-03
Add basic integer support.

ltratt:
Including support for +, -, *, and /. Note that SOM (like most Smalltalks) parses expressions such as "2 + 3 * 4" as "(2 + 3) * 4": it has no concept of operator precedence, treating all binary operations as left associative.

### DIFF
--- a/src/lib/compiler/ast.rs
+++ b/src/lib/compiler/ast.rs
@@ -27,6 +27,7 @@ pub struct Method {
 
 #[derive(Debug)]
 pub enum MethodName {
+    BinaryOp(Lexeme<StorageT>, Option<Lexeme<StorageT>>),
     Id(Lexeme<StorageT>),
     Keywords(Vec<(Lexeme<StorageT>, Lexeme<StorageT>)>),
 }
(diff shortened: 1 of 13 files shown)


src/lib/compiler/ast.rs
```diff
@@ -27,6 +27,7 @@ pub struct Method {
 
 #[derive(Debug)]
 pub enum MethodName {
+    BinaryOp(Lexeme<StorageT>, Option<Lexeme<StorageT>>),
```
vext01: How comes the RHS is optional here?

src/lib/compiler/ast.rs
```diff
@@ -46,11 +47,17 @@ pub enum Expr {
         id: Lexeme<StorageT>,
         expr: Box<Expr>,
     },
+    BinaryMsg {
```
vext01: What's the difference between `BinaryOp` and `BinaryMsg`? I guess the latter has something to do with calling?

src/lib/compiler/ast.rs
vext01: It's to do with left to right evaluation, isn't it? Does a `None` terminate a chain of binary ops?

ltratt: This one caught me out too yesterday: this is the method *name*, not the method call. So it allows method names like `+` instead of `+ argument`. I'm not quite sure why those are allowed yet, but I guess it'll become obvious at some point!

src/lib/compiler/ast.rs
ltratt: `BinaryOp` is a method name and `BinaryMsg` is a method call. [IIRC, the names are taken from the grammar, which was roughly taken from the original SOM parser.]

src/lib/compiler/ast.rs
vext01: OK, let's roll with it for now.

src/lib/compiler/ast.rs
vext01: ok

PR #33, opened 2019-09-05
Loops

ltratt:
This PR adds support for `while` loops. The main surprise to me is the `restart` primitive which in SOM restarts the "calling" function (in a sense) in https://github.com/softdevteam/yksom/commit/f0ffa3a807dc44c750ff4985854ae444bf22229b. I'm not sure if that's a normal Smalltalk thing or not, but the easiest way to think about what it does is to look at the `whileTrue` method definition.

Before we can do that, we need to enable basic integer comparisons (https://github.com/softdevteam/yksom/commit/af161b050cc92718edf8d8bca458d532fd23f946). We then finish with an "I'm an unthinking idiot" scoping fix (https://github.com/softdevteam/yksom/commit/fd61a14158b3f1e3534b1d7d15540e1b7c8bc8cc).


vext01:
So if I see it correctly, the loop condition is captured from the environment, like a closure, then the block is continually restarted until the condition doesn't hold. Is that right?

smarr:
The `Block>>#restart` primitive sets the instruction pointer back to zero. For example: https://github.com/SOM-st/som-java/blob/master/src/som/primitives/BlockPrimitives.java#L42-L44

There are other ways to implement it.
[Smalltalk 80 has jump bytecodes](http://www.mirandabanda.org/bluebook/bluebook_chapter28.html#JumpBytecodes28), TruffleSOM simply replaces the `#whileTrue:` AST node with one that loops, you can also simply do a recursive implementation of `#whileTrue:` (you probably want tail call optimization for that), and there're probably other solutions.

The main question is: what can you get away with without getting caught cheating.
Bytecodes are easily detected, and a violation of the language semantics. A `#restart` primitive is a conceptually simple solution, but may not be the one with the best performance.

Also, @vext01, the loop condition is explicitly a closure, it is a block, and more specifically it is `self` in the `#whileTrue:` method.

ltratt:
> TruffleSOM simply replaces the #whileTrue: AST node with one that loops

That's interesting. I was aiming for 100% source compatibility and no cheating. It would clearly be much faster for an interpreter to implement `whileTrue` internally, but it feels like that would be cheating to me. Is this because TruffleSOM can't implement `restart`?

smarr:
I do consider the AST replacement cheating, yes. And with AST replacement, `#reset` just gives an error.

In a Truffle-style AST interpreter, restart for the `#whileTrue` use case could be simply a control flow exception, I think. I didn't try, or compare whether that is equivalent in performance. It should be, in an ideal world.
Historically, the AST replacement was just the simpler thing to do. And at that time Graal struggled with a few compiler bugs that made my life hard (non-local returns were not used by any other language before). So, I think, I just chose the simples solution.

ltratt:
@smarr OK, understood, and thanks for the candour!

@vext01 Hopefully Stefan's initial explanation cleared up the role of `restart`?

vext01:
Do you need to squash anything?

ltratt:
I think we can merge as-is.

PR #34, opened 2019-09-07
Block parameters

ltratt:
This PR first implements some more integer comparisons (https://github.com/softdevteam/yksom/commit/3869e567c0e5e4fa5c63971caffb7738ad332fe9) which are mostly rote.

It then implements block parameters (https://github.com/softdevteam/yksom/commit/68ba59645a278e78d83896c5605ee4799f7d4ebe). The way SOM does this is a bit cack-handed. A block without parameters is an instance of `Block`, a block with 1 parameters is an instance of `Block2` and a block with 2 parameters is an instance of `Block3`. You can't have blocks with more than 2 parameters AFAIK. This support allows us to do the SOM equivalent of `for` loops.


smarr:
A little unfortunate the design, I agree.

But, support for arbitrary blocks would mean a change to the standard library (which I consider part of the language), so, while possible (https://github.com/charig/TruffleMATE/pull/31, https://github.com/smarr/SOMns/issues/94), I'd consider it undesirable as any other "language" change.

lang_tests/block10.som
```diff
@@ -0,0 +1,18 @@
...
+    4
+"
+
+block8 = (
+    f = (
+        self g: [ :x :y | ^ (x + y) ].
+        ^'b'
+    )
+
+    g: block = ( ^block value: 2 with: 2 )
```
smarr: I would recommend `^ x - y` and then here `block value: 10 with: 3`.
This way, you'd also test whether the arguments are in the correct order.

lang_tests/int6.som
```diff
@@ -14,5 +26,21 @@ int6 = (
         (1 = 1) asString println.
         (1 ~= 0) asString println.
         (1 ~= 1) asString println.
+
+        (1 < 1) asString println.
```
smarr: You could implement `println` on `Object` and avoid all the `asString` duplication (see the SOM core-lib).

lang_tests/block10.som
ltratt: That's a very good point. Fixed in https://github.com/softdevteam/yksom/pull/34/commits/8301841a0dafac232e5619856b203ab3a6ee44a2.

lang_tests/int6.som
ltratt: I agree that this is a bit unfortunate at the moment. I'll fix it soon and aim to remove all the unnecessary `asString` calls, but probably not in this PR.

lang_tests/block10.som
```diff
@@ -0,0 +1,18 @@
...
+    7
+"
+
+block8 = (
+    f = (
+        self g: [ :x :y | ^ (x - y) ].
+        ^'b'
+    )
+
+    g: block = ( ^block value: 10 with: 3 )
```
vext01: Do I see it correctly that this is how you invoke the block you've defined above? The first argument is implicitly named `value` and the second `with`?

vext01: Later code confirms this. Unusual...

vext01:
Please squash

lang_tests/block10.som
smarr: How do you represent names for arguments in a language that requires to name them explicitly?

The common choices are:
 - `#value:`, `#value:value:`, `#value:value:value:`, etc
 - `#value:`, `#value:with:`, `#value:with:with:`, etc

You'll see similar patterns for other APIs that require multiple arguments.

ltratt:
Squashed.

lang_tests/block10.som
vext01: It's not the syntax that surprised me. Blocks seem like a general language construct, but they can only have 0, 1 or 2 parameters and with very specific names.

Maybe I misunderstood...

ltratt: The parameters *inside* the block can have arbitrary names, but the way you call them is with fixed names. For those of us who aren't used to thinking in a Smalltalk-y way, this is pretty weird, I admit.

smarr: The restriction of the number of parameters is unfortunate, and specific to SOM.
The links in my first comment show how one can get to arbitrary number of arguments.
Though, that does not fix the "ugly" handling of large number of arguments. You'll still end up with `block value: 1 with: 2 with: 3 with: 4 with: 5 with: 6 with: 7` and so on.
I think Pharo/Squeak have explicit methods for up to about 7 arguments.
Afterwards, one can use `block valueWithArguments: {1. 2. 3. 4. 5. 6. 7. 8. 9.}` in Pharo/Squeak.

This is one big drawback of keyword messages. And there have been probably a million proposals on how to improve on that in Smalltalk over the decades :)

For the debate of names of parameters.
Well, I think there is a misunderstanding. The keyword names of methods do not specify the name of the argument. They are just a guide so that the resulting "sentence" becomes readable.
And there's no difference to the handling of names of arguments compared to other position-based argument passing languages.
In Java, you can also change the name of arguments in methods of subclasses. You just can't change their order/position.

So, the keyword parts constitute the method name. Under the hood it looks like for other languages: `block value: 1 with: 2 with: 3` becomes `block.value:with:with:(1, 2, 3)`, as it would be in Java or JavaScript.

PR #36, opened 2019-09-12
Implement arbitrary sized integers.

ltratt:
This allows arithmetic operations to operate on numbers that are as large as you have RAM for (i.e. you're not restricted to working on integers in machine words). We use the `num_bigint` library to represent such arbitrary sized integers.

Arbitrary sized integers are not an efficient way of representing small numbers, so we now have three ways of representing integers:

  * Tagged fixed-sized integers (i.e. an isize minus 3 bits)
  * Boxed fixed-sized integers (boxed isizes)
  * Boxed arbitrary-sized integers (boxed BigInts)

For any given number we always try to store it as the most efficient representation (i.e. if "x" is a BigInt and you do "x-x" you'll get back a tagged integer representing 0). This is completely transparent to the user: SOM simply exposes an "integer" type and doesn't differentiate between fixed-sized and arbitrary-sized integers.


src/lib/vm/objects.rs
```diff
@@ -80,6 +82,14 @@ pub trait Obj: Debug + abgc::GcLayout {
         Err(Box::new(VMError::CantRepresentAsUsize))
     }
 
+    /// If possible, return this `Obj` as a `BigInt`.
+    ///
+    /// XXX This API is clearly inefficient in the sense that it forces a `ArbInt` to clone an
```
ptersilie: Are we gonna leave that TODO in? That makes me wonder: do we have a code standard for this, i.e. whether we use "XXX" or "TODO"?

ptersilie:
Just a minor comment/question.

src/lib/vm/objects.rs
ltratt: I take `XXX` to mean "this isn't really the right way of doing things and should be improved" whereas `TODO` is more along the lines of "extra things that would be nice". So I think we should probably leave this one as-is?

vext01: FWIW, I consider `XXX`, `FIXME` and `TODO` pretty much the same.

I prefer `XXX` as it's easier to type, but the Rust world seems to have adopted `FIXME` (using `XXX` in the compiler will actually crash the build due to a lint check).

ltratt: I don't feel stongly about this, so let's go with `FIXME`.

ltratt: Fixed in the latest commit.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #37, opened 2019-09-19
Tidy up numeric operations

ltratt:
Our numeric operations (add, sub, equals, not equals etc.) had gotten into a bit of a mess. This PR fixes the two major problems:

 1) We have 3 ways of representing integers (tagged integers, boxed (machine word) integers, and arbitrary sized integers). We want to make this transparent to the user while also performing operations in the most efficient manner possible. Previously we had auto-casting methods on `Obj`, but these are easy to abuse, and to end up with inefficient (and, sometimes, incorrect) casts. We now do much more of the work in `Val` (https://github.com/softdevteam/yksom/commit/8aa717a8bbd512aadc532ed252ed295935f16b7f).

 2) This then makes clear that the `Obj` trait is something that people *implementing* types for yksom need to care about, but people *calling* things shouldn't worry about directly. Previously we forced people to regularly use the `tobj` method to box a possibly non-boxed value: now we duplicate every function in `Obj` in `Val` and tell people to use the latter instead. This simplifies all code outside of `Val`, and also makes things faster, because we can make full use of tagged integers (https://github.com/softdevteam/yksom/commit/718b85a0ad14e7b95a99ae780c5f2686006f3f56).


src/lib/vm/objects.rs
```diff
@@ -541,6 +431,18 @@ impl Int {
...
+
+    pub fn as_isize(&self) -> isize {
+        self.val
+    }
+
+    pub fn as_usize(&self) -> Option<usize> {
+        if self.val > 0 {
+            Some(self.val as usize)
+        } else {
+            None
```
ptersilie: Why can't we cast a value of 0 into a usize here?

ltratt: Good spot! Fixed in https://github.com/softdevteam/yksom/pull/37/commits/9574d365a42fcfecb03e58a33c4d078d0407602c.

ptersilie: Hah, I didn't think this was actually a bug, just something about the internals I don't quite understand.

ptersilie:
Rest looks good. Please squash!

src/lib/vm/objects.rs
ltratt: That's why reviews are valuable -- sometimes the obvious thing gets forgotten!

ltratt:
Squashed.

lang_tests/int10.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: NotANumber { got: String_ }
+"
+
+int10 = (
+    run = (
+        0 = 'a'
```
smarr: I don't understand why this would error.

The expected behavior is:

```
SOM Shell.
---> 0 = 'a'
it = false
```
From https://som-st.github.io/

lang_tests/int11.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: NotANumber { got: String_ }
+"
+
+int11 = (
+    run = (
+        0 ~= 'a'
```
smarr: Same thing with all of these tests:

```
---> 0 ~= 'a'
it = true
```

ltratt: You're right. I must have mixed up the semantics in my head -- mea culpa!

@ptersilie I should fix these before we merge this PR.

lang_tests/int10.som
ptersilie: Is it because of the 0  to usize change? Or did this error already before?

ltratt: I'm afraid this is solely due to my stupidity: I didn't check SOM's semantics carefully enough.

lang_tests/int11.som
ptersilie: Okay, I will wait for the fixes then!

lang_tests/int10.som
smarr: Would it help if I somehow try to get your tests runnable on another SOM?

ltratt: In this case, nothing would have prevented me being an idiot. I think I know what I did: I tried `>` in Java SOM, saw that it triggered an error, and generalised from there.

At some point I will get (most) of the tests running on another SOM. The ones that won't work are those that check stderr: each SOM fails in different ways on error. But the ones that succeed should only require a small tweak to `lang_tests/run.rs` (see https://crates.io/crates/lang_tester).

lang_tests/int11.som
ltratt: Should be fixed in https://github.com/softdevteam/yksom/pull/37/commits/4fb753c4bea73f7bde80eba3b83c598aeb5cfa8c.

ptersilie:
Looks good. Want to squash or leave it?

ltratt:
Squashed.

PR #39, opened 2019-09-19
Generate cargo doc information on gh-pages for each PR.

ltratt:
Shamelessly copied from grmtools. I don't think I can test this before it's merged, but I might be very wrong. Any thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,11 @@ before_script:
 script:
     - cargo +nightly fmt --all -- --check
     - cargo test
+    - cargo doc --no-deps
+deploy:
+    provider: pages
+    skip-cleanup: true
+    github-token: $GITHUB_TOKEN
+    local-dir: target/doc
+    on:
+        branch: master


ltratt:
Sausages. That README commit shouldn't be there. I'll remove that before merging!

.travis.yml
```diff
@@ -7,3 +7,11 @@ before_script:
 script:
     - cargo +nightly fmt --all -- --check
     - cargo test
+    - cargo doc --no-deps
+deploy:
+    provider: pages
+    skip-cleanup: true
+    github-token: $GITHUB_TOKEN
+    local-dir: target/doc
```
ptersilie: Why is this `target/doc` here, but `doc/release` in grmtools?

ltratt: In `grmtools`, we're using `mdbook`, but here we're using `cargo doc` which dumps stuff in `target/doc`. At least I think that's right, but I'm hoping you can point if/where I'm wrong!

ptersilie:
It could be that if you change `master` to the branch it will build it on the branch. But I could be totally wrong and travis only does "deploy" on the master branch. There may be something in the travis docs.

.travis.yml
ptersilie: Sounds plausible.

ltratt:
I have added a GH token to travis, and I've checked the documentation and I *think* things are OK. Sholud I squash and we try merging and cross our fingers?

ptersilie:
Let's do it!

ltratt:
I've removed the unintended commit, so I think this is ready to merge.

PR #40, opened 2019-09-20
Change the name on gh-pages of the internal API documentation.

ltratt:
Instead of being `yksom/yksom`, this will make it appear as `yksom/api` (leaving us wiggle room if we want to put non-API docs on gh-pages later).


ptersilie:
Looks good.

Issue #38, opened 2019-09-19
More descriptive readme


bjorn3:
What is SOM?

ltratt:
Point taken -- I agree that this has reached a point where fleshing out the `README` a little bit makes sense.

ltratt:
Fixed in https://github.com/softdevteam/yksom/pull/42.

PR #42, opened 2019-09-21
Flesh out the README a little bit.

ltratt:
Needs #41 to be merged first. Addresses #38.

This PR fleshes out the README a bit. Since the README now links to the internal API docs, it also makes sense to flesh those out a little bit too. A lot more *could* be done here, but this is a start, at least.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,14 @@
 # ykSOM
 
-A Yorick SOM interpreter.
+A [SOM](http://som-st.github.io/) VM in Rust. SOM is a cut-down Smalltalk-like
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,3 +1,14 @@
 # ykSOM
 
-A Yorick SOM interpreter.
+A [SOM](http://som-st.github.io/) VM in Rust. SOM is a cut-down Smalltalk-like
```
bjorn3:
```suggestion
A [SOM](https://som-st.github.io/) VM in Rust. SOM is a cut-down Smalltalk-like
```

PR #43, opened 2019-09-22
Add shift left.

ltratt:
This will help make writing some tests involving big ints a little easier, hence why I'm doing it a bit sooner than expected.

@smarr I think `lang_tests/som22.som` shows up a bug in Java SOM (it seems to treat `(1 << 200) << (1 << 150)` as `(1 << 200)`). I'd be grateful for your thoughts on that!


smarr:
>  bug in Java SOM (it seems to treat (1 << 200) << (1 << 150) as (1 << 200))

This is more complicated, it's not just "laziness"...

The current implementation likely explains the behavior:
https://github.com/SOM-st/som-java/blob/3fab8ac7f34cedbbe15719fc607bceef25343d59/src/som/vmobjects/SInteger.java#L253

We just fall back to the `<<` operator of Java on a `long` type.

From a language design perspective, I think I have convinced myself to some degree, that `<<` and other bit operations have no place on integer types.
They are a hack, a leak of the underlying representation, and thus, should not exist.

A while ago, I pondered about this in a little more detail here:
https://github.com/smarr/SOMns/issues/218

For SOM, I don't really know what the best course of action is.
It is great to keep the number of concepts minimal and the number of standard library classes with it. On the other hand, there is a conceptual problem here...

At the moment wondering whether it would be a good compromise to declare all bit operations special, to restrict them as operating on the range of machine-word-sized two-complement integers only. This avoids introducing an extra concept, but defines the semantics in a way that is useful for the current use cases: benchmarks mostly.

So yeah, at the moment, there isn't a spec, and benchmarks are the main use case, which biases all implementations to doing things that are fast. Not an ideal situation.

ltratt:
>The current implementation likely explains the behavior:
https://github.com/SOM-st/som-java/blob/3fab8ac7f34cedbbe15719fc607bceef25343d59/src/som/vmobjects/SInteger.java#L253
>
>We just fall back to the << operator of Java on a long type.

I'm not sure that's entirely true: there is code to widen in that function it seems? Though I'm not entirely sure when it triggers.

smarr:
Hm, you're right, I ignored that part.
Looking at [IntegerTest](https://github.com/SOM-st/SOM/blob/master/TestSuite/IntegerTest.som) looks like it should be widening.

But, now reading the implementation, the shift argument is only supported to be in the integer range, which your argument isn't.

So, yeah, I guess widening is the way to go for the moment.

ltratt:
I think you're right: I think Java SOM hasn't implemented all the reasonable cases of widening (I don't know about other SOMs).

ltratt:
@ptersilie I need to make some safety improvements before merging this one, so give me a little bit of time.

ltratt:
Fixed (hopefully). https://github.com/softdevteam/yksom/pull/43/commits/2a0e14d4db90aef75e981c3027ef3b455ca4c0c9 fixes a problem that should have been addressed in the original PR, so will need squashing away before merging.

src/lib/vm/val.rs
```diff
@@ -366,6 +367,36 @@ impl Val {
         self.tobj(vm).unwrap().div(vm, other)
     }
 
+    /// Produce a new `Val` which shifts `self` `other` bits to the left.
+    pub fn shl(&self, vm: &VM, other: Val) -> ValResult {
+        if let Some(lhs) = self.as_isize(vm) {
+            if let Some(rhs) = other.as_isize(vm) {
+                if rhs < 0 {
+                    return ValResult::from_vmerror(VMError::NegativeShift);
+                } else {
```
ptersilie: Would it be worthwhile to check if `rhs` is zero and just return a copy of self then rather than doing all these other checks below?

src/lib/vm/val.rs
```diff
@@ -376,15 +376,21 @@ impl Val {
...
                     let rhs_i =
                         rtry!(u32::try_from(rhs).map_err(|_| Box::new(VMError::ShiftTooBig)));
-                    match lhs.checked_shl(rhs_i) {
-                        Some(i) => return Val::from_isize(vm, i),
-                        None => {
-                            return ArbInt::new(
-                                vm,
-                                BigInt::from_isize(lhs).unwrap() << usize::try_from(rhs_i).unwrap(),
-                            )
+                    if let Some(i) = lhs.checked_shl(rhs_i) {
```
ptersilie: Why is this not caught by `checked_shl`. I thought it returns `None` if it overflows?

src/lib/vm/val.rs
ltratt: This is changed in one of the later commits, but if it hadn't have been, I probably would have suggested keeping things as-is. I doubt people `shl` by `0` very often, and so the extra check probably doesn't win us enough to pay off. This is all a guess, of course, without benchmarks!

src/lib/vm/val.rs
ltratt: It returns `None` if it overflows 64 bits, but half of the values of an `isize` are negative. Put another way, it depends whether you view the underlying 64 bits as an `isize` or a `usize`. The underlying shift is correct either way, but if you view things as an `isize` then you get surprising results in cases where the left-most bit has now become 1.

src/lib/vm/val.rs
ltratt: Oops, I thought this was on the original commit, not the fix commit. Mea culpa! My comment, I think, still stands: I doubt the extra check is worth it (at least, not without benchmarks).

ptersilie: Makes sense.

src/lib/vm/val.rs
ptersilie: I see.

ptersilie:
Travis had an odd error, so I kicked it again. If it succeeds this would be ready for squashing.

ltratt:
Sounds reasonable!

ltratt:
@ptersilie If you're OK with this, I should squash those two fixes away.

ptersilie:
Please do.

ltratt:
Semi-squashed: I realised that https://github.com/softdevteam/yksom/pull/43/commits/baa1942d4893f76ce8f803e124fc82dd573a70ad was best left as a separate commit, as this PR exposed the bug, but didn't cause it.

PR #47, opened 2019-09-27
Implement basic support for doubles.

ltratt:
This implements operations such as `+-*` (though note that Doubles don't implement `/` in SOM!) but not comparisons (as we really need to improve some of our other support before it's worth fixing doubles) and many other necessary operations. However, it does largely integrate Doubles into the internal type system.

Note that this PR starts to diverge from SOM for minor reasons over formatting of doubles (https://github.com/SOM-st/SOM/issues/22) and inconsistencies in Java SOM (at least) (https://github.com/SOM-st/SOM/issues/20). Hopefully we will be able to fix that divergence at some point.


ltratt:
bors try

PR #48, opened 2019-09-28
Downcasting on tagged integers needs to be handled.

ltratt:
This is one of those "what was I thinking?!" moments where it's hard to imagine how I got so confused that I made the mistake this commit fixes.

yksom, correctly, stops you trying to downcast a `Val` to an `Int`. However, the `*downcast` functions took that too far by also trying to stop you checking whether a run-time `Val` that happened to be an `Int` could be downcast or not. If you did, you got (in debug mode) an assertion failure or (in runtime mode) undefined behaviour. This simple commit rectifies this mistake by explicitly checking for tagged integers when downcasting.

### DIFF
--- a/src/lib/vm/val.rs
+++ b/src/lib/vm/val.rs
@@ -131,28 +131,30 @@ impl Val {
         &self,
         _: &VM,
     ) -> Result<&T, Box<VMError>> {
-        debug_assert_eq!(self.valkind(), ValKind::GCBOX);
-        debug_assert_eq!(ValKind::GCBOX as usize, 0);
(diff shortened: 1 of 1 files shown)


ltratt:
There is now a test (it's so easy that I can't believe I didn't add it earlier and spot the problem earlier!) https://github.com/softdevteam/yksom/pull/48/files#diff-b13ca67b1dcc5e9eaf228b5ebd79a5b3R722

ptersilie:
Oh, didn't see this was a test. My bad.

ltratt:
Not your fault! It's pretty subtle that it tests this explicitly.

PR #49, opened 2019-09-28
Have buildbot test release mode too.

ltratt:
yksom's debug and release mode are somewhat different, so it's worth taking the (smallish) extra time to test both.


ltratt:
bors try

PR #50, opened 2019-09-30
Split apart objects.rs

ltratt:
`objects.rs` was becoming a monster (1000 lines) and is only going to grow. This PR minimally splits things apart (https://github.com/softdevteam/yksom/commit/16009256db15a18199d1dea44fca217cd2a31e97) and then makes a few minor fix-ups.

I'm not totally sure if this is the best way of splitting it up, so I'm open to suggestions!


src/lib/vm/objects/integers.rs
```diff
@@ -7,6 +7,10 @@
 // at your option. This file may not be copied, modified, or distributed except according to those
 // terms.
 
+//! yksom has three ways of representing integers: as tagged integer (inside a `Val`); boxed
```
ptersilie: Why are tagged integers (`Val`) not inside this module as well?

ltratt: `Val` is the union of tagged integers and boxed objects, so it's a very separate thing (in its own file `val.rs`). In a sense, `Val` stands outside the normal object system, if that makes sense.

ptersilie: It does. Thanks for the clarification.

ptersilie:
Do you want to squash anything or is this ready for merging as is?

ltratt:
Merge as-is please!

PR #53, opened 2019-10-07
Simplify and improve the core interpreter loop

ltratt:
The core of the VM is the interpreter loop. I had previously made this very generic (and inefficient) while I tried to understand SOM's semantics better. For example, I assumed that SOM had continuations and pushed function stacks into frames for simplicity -- but my assumption was wrong.

This PR gradually strips several of these assumptions away, making the code easier to understand and faster (mostly by avoiding unnecessary intermediate memory allocations). It starts by moving to a single SOM stack (https://github.com/softdevteam/yksom/commit/d5b8d20a4905086c9b118dfa844c8c5874d7b75a), and then simplifies primitives (which only make sense in the current function frame https://github.com/softdevteam/yksom/commit/245746e796e0a8d04c99d6004518f550c7ba4099), before a series of commits simplifies message sending (notably https://github.com/softdevteam/yksom/commit/2adb6e2ec0598d7ad17c1f00e2babae38481cc56 and https://github.com/softdevteam/yksom/commit/c6cc0656297430573bba78d4fbf2c63e690a91d7).

Each commit in this PR is correct and passes the tests, but it might not always be easy to see where they're going. It might be easiest to start by looking at `core.rs` [before](https://github.com/softdevteam/yksom/blob/b4eb616022dab8da5ef92d5e1e1f6813728db9fa/src/lib/vm/core.rs) and [after](https://github.com/ltratt/yksom/blob/one_stack/src/lib/vm/core.rs) this commit and then working through the commits individually, though obviously that's up to you!

At the moment there aren't a vast number of benchmarks we can run, and those we can only tell us a limited amount about yksom. That said, this simple benchmark:

```
while_bench = (
    run = ( | x |
        x := 1000000.
        [x ~= 0] whileTrue: [
            x := x - 1.
        ].
    )
)
```

is sped up a smidgen over 20% by this PR.


ltratt:
bors try

src/lib/vm/core.rs
```diff
@@ -432,23 +438,64 @@ impl VM {
...
     }
+
+    fn stack_len(&self) -> usize {
+        unsafe { &*self.stack.get() }.len()
+    }
+
+    /// Returns the top-most value of the stack without removing it. If the stack is empty, calling
+    /// this function will lead to undefined behaviour.
+    fn stack_peek(&self) -> Val {
+        // Since we know that there will be at least one element in the stack, we can use our own
```
ptersilie: This comment sort-of makes sense here too (even if we don't really pop), but I just wanted to make sure you meant to copy paste the same comment here.

ltratt: Yep, it's deliberate.

src/lib/vm/core.rs
```diff
@@ -58,8 +58,6 @@ pub enum VMError {
     },
```
ptersilie: Typos in commit message:
- Missing quote around `restart bytecode`
- made things a **bit** too generic

src/lib/vm/core.rs
```diff
@@ -73,6 +71,28 @@ pub enum VMError {
     UnknownMethod(String),
 }
 
+#[derive(Debug)]
+/// The (internal) result of a SOM send.
+enum SendReturn {
+    /// A closure wants to perform a return *n* frames up the call stack with a.
```
ptersilie: What is `a`?

ltratt: Oops! Fixed in https://github.com/softdevteam/yksom/pull/53/commits/e0af0d21ff2ce9efeea30a446a1832f8b37b7583.

ltratt:
If that new fix works for you, then I can squash and fix the commit messages at the same time.

ptersilie:
Please do!

ltratt:
Squashed and fixed.

PR #55, opened 2019-10-08
More comparisons

ltratt:
This PR adds more comparisons to yksom: reference equality (https://github.com/softdevteam/yksom/commit/1dca2bc9024d03becd7d92a4c7abf1a372765c03) and various comparisons for doubles (https://github.com/softdevteam/yksom/commit/82d35c5238038d79c7e99498d0988e4359a06dca). Notice that the former commit means that we diverge slightly from SOM's current semantics (see https://github.com/SOM-st/SOM/issues/23) although I think it's fair to say that @smarr thinks the approach we're taking in this PR is the right one (and if I've got that wrong, hopefully he'll shout!).

### DIFF
--- a/lang_tests/double10.som
+++ b/lang_tests/double10.som
@@ -1,0 +1,46 @@
+"
+VM:
+  status: success
+  stdout:
+    false
+    true
+    true
+    true
+    false
+    true
+    false
+    false
+    false
+    true
+    true
+    false
+    true
+    true
+    true
+    false
+"
+
+double10 = (
+    run = (
+        (1.0 = 0) println.
+        (1.0 = 1) println.
+        (1.0 ~= 0) println.
+        (1.0 ~= 1) println.
+
+        (1.0 < 1) println.
+        (1.0 < 2) println.
+        (2.0 < 0) println.
+
+        (1.0 > 1) println.
+        (1.0 > 2) println.
+        (2.0 > 0) println.
+
+        (1.0 >= 1) println.
+        (1.0 >= 2) println.
+        (2.0 >= 0) println.
+
+        (1.0 <= 1) println.
+        (1.0 <= 2) println.
+        (2.0 <= 0) println.
+    )
+)
--- a/lang_tests/obj1.som
+++ b/lang_tests/obj1.som
@@ -1,0 +1,29 @@
+"
+VM:
+  status: success
+  stdout:
+    true
+    false
+    false
+    true
+    true
+    true
+    false
+    false
+    true
+"
+
+obj1 = (
+    run = (
+        (1 == 1) println.
+        (1 <> 1) println.
+        (1 ~= 1) println.
+        (1 <> 2) println.
+        (1 ~= 2) println.
+
+        (self == self) println.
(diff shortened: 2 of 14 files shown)


lang_tests/double10.som
```diff
@@ -0,0 +1,46 @@
...
+    false
+    true
+    true
+    true
+    false
+"
+
+double10 = (
+    run = (
+        (1.0 = 0) println.
```
ptersilie: Shouldn't this be `==`?

lang_tests/double10.som
```diff
@@ -0,0 +1,46 @@
...
+
+        (1.0 < 1) println.
+        (1.0 < 2) println.
+        (2.0 < 0) println.
+
+        (1.0 > 1) println.
+        (1.0 > 2) println.
+        (2.0 > 0) println.
+
+        (1.0 >= 1) println.
```
ptersilie: Wait, so in SOM, `1.0 == 0` is false, `1.0 > 1` is false, but `1.0 >= 0` is true?

smarr: `==` is very different from the other two, and unrelated.

`1.0 > 1` and `1 > 1` should yield both `false`, no? (minus issues with number representations for floating points, which hopefully do not matter for 1-like values, but will crop up with int-like values > 53-or-something bits).

And `1.0 >= 1` and `1 >= 1` should both be true in and ideal world (minus floating point issues)

lang_tests/double10.som
ltratt: SOM is a little unusual (to our eyes) in this regadr. `=` means "check the values for equality". `==` means (roughly) "check the data pointer for equality" (though for integers this is overridden -- Python effectively does something similar).

ltratt:
@smarr Just to confirm, do all the checks make sense to you (at least in yksom)?

smarr:
@ltratt sorry, I didn't check in detail the true/false values. The test structure makes it hard for me to match up tested expression and expected value.
But, the general direction looked about right. Any difference, you'll notice soon enough when you try to run the SOM test suite :)

ltratt:
I compare Java SOM and yksom's output if that's any help. There are differences in `lang_test/obj2.som`, in particular.

ltratt:
@smarr OK, thanks! yksom, of course, aims to be an ideal world, a veritable nirvana, a stronghold of God's will on Earth. Except bugs, of course ;)

lang_tests/double10.som
ptersilie: I see. Thanks for the explanation.

lang_tests/double10.som
ptersilie: Ah thanks! With this comment and the @ltratt's comment above, this makes a lot more sense now.

ptersilie:
Okay, I guess this is ready to merge then and I can tell bors the good news?

ltratt:
Yes please!

vext01:
Sorry, my CI changes yesterday broke this. I've applied the fix, but I just need a:

bors r+

PR #56, opened 2019-10-12
Add bitXor primitive.

ummarikar:
SOM currently only defines bitXor's between two Ints, so we throw an error in other circumstances.


ltratt:
At first glance this looks good, but here are a few comments:

* This PR is missing a meaningful title and description (both are absolutely necessary).

* This should almost certainly be 1 commit. Commits need properly capitalised titles (and full stops at the end) and a brief message explaining what they do. [When you have a PR that's 1 commit big -- as this PR probably should be -- the PR and commit message will almost certainly be identical, so that cuts the necessary work down!]

* We should think about what "[<int | bigint>] ^ <double>" should do. What does it do in Java SOM? I guess it just crashes?

To fix the first point, you can just edit the PR directly. To fix the second point, you might as well squash and force push now (you should only force push when the reviewer asks, but I'm asking, so that's OK!). And the third point might need a bit of discussion because I'm not sure what the right answer is.

ltratt:
And so that you can see what bors and buildbot will do when this PR is ready: bors try.

ltratt:
bors try

ltratt:
[If you click on the red/orange/green dot next to "verified", and then click on "details", you can see buildbot testing this PR.]

ltratt:
That error is because you probably forgot to run `cargo fmt` before committing and raising the PR. You *might* need to install the nightly version of `cargo-fmt` in order to be able to format this project, although you might not (I can't remember anymore!).

lang_tests/double11.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: CantRepresentAsDouble
```
smarr: Hm, this error seems confusing.
I'd expect something like an incompatible type error or a InvalidArgument error.

----
[Distraction]

Perhaps more confusingly, why isn't just the bit pattern of the double used?
We just use the two-complement bit pattern that is the assumed representation of the int as well, so, seems inconsistent to not simply use the bit representation of the IEEE 754 floating point number.
(There are likely very good reasons why this would not be useful in practice, but it seems like a conceptual inconsistency. And sorry for the distraction here, this is just another example why https://github.com/SOM-st/SOM/issues/21 needs more thoughts.)

lang_tests/int25.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: NotANumber { got: String_ }
```
smarr: This error could also be confusing because `bitXor:` doesn't accept just any kind of number.

lang_tests/double11.som
ltratt: I think this is a good question, and I don't know what SOM's intended semantics are. There are a number of reasonable answers here. Any thoughts?

smarr: At the moment, there are unfortunately no 'intended' semantics. Though, there are 'required' semantics, for the benchmarks, and they only care about the first 32-bits of integers, if I recall correctly.

ltratt: As the *de facto* SOM maintainer, I would welcome your thoughts on what your preferred semantics might be :) [If you don't have an opinion then, unless the benchmarks `xor` doubles, I'd be inclined to error at the moment.]

smarr: Yeah, no, current semantics should error. That's fine. I was just not happy with the `CantRepresentAsDouble`, which seems confusing to me.

The rest, I tried to mark as [Distraction] and it's more of a far off discussion.

ltratt: Point taken!

@ummarikar I suggest that we a) only allow integers (small or big) to be bitxor'd and thus b) throw a `TypeError` when someone passes something else (where `expected` will be `Integer`).

ummarikar: Okay will change this to a type error

lang_tests/int25.som
ummarikar: Understood will change this to a `TypeError`

ummarikar:
bors try

ltratt:
@ummarikar I'm going to give you permission to do a try but please make sure the only thing you do is to do a try (you could in theory tell bors to merge the PR, so please don't do that!).

bors d+

ummarikar:
bors try

lang_tests/int24.som
```diff
@@ -0,0 +1,16 @@
...
+  status: success
+  stdout:
+    1
+    1606938044258990275541962092341162602522202993782792835301377
+    176763184868488930309615830157527886277442329316107211883151360
+"
+
+int24 = (
+    run = (
+        (5 bitXor: 4) println.
```
ltratt: OK, so one other test we need here is `smallint bitXor: bigint`. Once you've added that, I think we'll have the complete set of possibilities.

ummarikar:
bors try

ltratt:
Please squash.

ltratt:
[When you've squashed you need to say something like "Squashed" so that I know it's happened.]

ummarikar:
bors try

ummarikar:
> [When you've squashed you need to say something like "Squashed" so that I know it's happened.]

Will remember this for next time!

ltratt:
OK, so one comment, for which you'll need to do a forced push: commit messages need to have a short title (80 characters or less except in rare circumstances) and a more detailed body. So this commit might have a title "Add bitXOr primitive." and a body saying something like "SOM currently only defines bitXOr's between two Ints, so we throw an error in other circumstances". Note that the body also needs to be word wrapped to 80 chars.

ltratt:
[No need to do a `try` after that push. Assuming it's OK, bors will do a full check before merging this into master.]

ummarikar:
Hi @ltratt, updated commit message and force pushed.

ltratt:
Thanks!

bors r+

ltratt:
OK, your first PR is in! You'll probably want to delete the branch from your fork first, and then you can move on to the next primitive!

PR #57, opened 2019-10-17
Add sqrt primitive.

ummarikar:
This adds the sqrt primitive to yksom for Int, BigInt and Double.

### DIFF
--- a/lang_tests/int29.som
+++ b/lang_tests/int29.som
@@ -1,0 +1,18 @@
+"
+VM:
+  status: success
+  stdout:
+    Integer
+    Integer
+    Double
+    Integer
+"
+
+int29 = (
+    run = (
+        (4 sqrt class) println.
+        ((1 << 200) sqrt class) println.
+        (3 sqrt class) println.
+        ((1 << 199) sqrt class) println.
+    )
+)
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -41,6 +41,8 @@ pub enum VMError {
     /// A value which can't be represented in an `usize`.
     CantRepresentAsUsize,
     DivisionByZero,
+    /// A value which is mathematically undefined.
+    DomainError,
     /// The VM is trying to exit.
     Exit,
     /// Tried to perform a `Val::downcast` operation on a non-boxed `Val`. Note that `expected`
@@ -463,6 +465,10 @@ impl VM {
             }
             Primitive::Shl => {
                 self.stack_push(stry!(rcv.shl(self, self.stack_pop())));
+                SendReturn::Val
+            }
+            Primitive::Sqrt => {
+                self.stack_push(stry!(rcv.sqrt(self)));
                 SendReturn::Val
             }
             Primitive::Sub => {
--- a/src/lib/vm/objects/double.rs
+++ b/src/lib/vm/objects/double.rs
@@ -91,6 +91,10 @@ impl Obj for Double {
         }
     }
 
+    fn sqrt(&self, vm: &VM) -> Result<Val, Box<VMError>> {
+        Ok(Double::new(vm, self.val.sqrt()))
+    }
+
     fn ref_equals(&self, vm: &VM, other: Val) -> Result<Val, Box<VMError>> {
         let b = if let Some(rhs) = other.try_downcast::<Double>(vm) {
             self.val == rhs.double()
--- a/src/lib/vm/objects/integers.rs
+++ b/src/lib/vm/objects/integers.rs
@@ -17,7 +17,7 @@ use std::convert::TryFrom;
 
 use abgc_derive::GcLayout;
 use num_bigint::BigInt;
-use num_traits::{FromPrimitive, ToPrimitive};
+use num_traits::{FromPrimitive, ToPrimitive, Zero};
 
 use crate::vm::{
     core::{VMError, VM},
@@ -152,6 +152,14 @@ impl Obj for ArbInt {
             Err(Box::new(VMError::NotANumber {
                 got: other.dyn_objtype(vm),
             }))
+        }
+    }
+
+    fn sqrt(&self, vm: &VM) -> Result<Val, Box<VMError>> {
+        if self.val < Zero::zero() {
+            Err(Box::new(VMError::DomainError))
+        } else {
+            ArbInt::new(vm, self.val.sqrt())
         }
     }
 
--- a/src/lib/vm/val.rs
+++ b/src/lib/vm/val.rs
@@ -420,6 +420,23 @@ impl Val {
         self.tobj(vm).unwrap().shl(vm, other)
     }
 
+    /// Produces a new `Val` which is the square root of this.
+    pub fn sqrt(&self, vm: &VM) -> Result<Val, Box<VMError>> {
+        if let Some(lhs) = self.as_isize(vm) {
+            if lhs < 0 {
+                return Err(Box::new(VMError::DomainError));
+            } else {
+                let result = (lhs as f64).sqrt();
+                if result.round() == result {
+                    return Val::from_isize(vm, result as isize);
+                } else {
+                    return Ok(Double::new(vm, result));
+                }
+            }
+        }
+        self.tobj(vm).unwrap().sqrt(vm)
+    }
+
     pub fn to_strval(&self, vm: &VM) -> Result<Val, Box<VMError>> {
         debug_assert!(!self.is_illegal());
         match self.valkind() {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ itertools = "0.8"
 lrlex = "0.4"
 lrpar = "0.4"
 num-bigint = "0.2"
+num-integer = "0.1"
 num_enum = "0.3"
 num-traits = "0.2"
 natrob = { git="https://github.com/softdevteam/natrob", features=["abgc"] }
--- a/lang_tests/double12.som
+++ b/lang_tests/double12.som
@@ -1,0 +1,14 @@
+"
+VM:
+  status: success
+  stdout:
+    1.1
+    NaN
+"
+
+double12 = (
+    run = (
+        (1.21 sqrt) println.
+        (-1.1 sqrt) println.
+    )
+)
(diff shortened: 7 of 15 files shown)


ummarikar:
I require some guidance. I have put the necessary tests for positive numbers but I am unsure of what to do with negative numbers. In Java SOM, sqrt of negative numbers returns "NaN" and I am not sure what the equivalent of that is in yksom after some initial searching in the codebase. Would appreciate some help!

src/lib/vm/val.rs
```diff
@@ -420,6 +421,14 @@ impl Val {
         self.tobj(vm).unwrap().shl(vm, other)
     }
 
+    /// Produce new `Val` which is the square root of this.
```
ltratt: "Produces a new `Val`..."

ltratt:
> In Java SOM, sqrt of negative numbers returns "NaN" and I am not sure what the equivalent of that is in yksom after some initial searching in the codebase.

It should probably either produce `NaN` ("Not A Number") or throw an error. @smarr any thoughts on which? I'm inclined towards the latter, in part because `sqrt` in C throws an error (rather than producing `NaN`).

ltratt:
OK, having taken a quick look at this, I think the correct thing to do (in IEEE 574) is to return NaN, so let's do that (which, conveniently, is what Rust's `sqrt` does anyway) and add some appropriate tests.

ummarikar:
Just implemented NaN for negative square roots. Let me know if you want me to squash.

src/lib/vm/objects/double.rs
```diff
@@ -92,7 +92,11 @@ impl Obj for Double {
     }
 
     fn sqrt(&self, vm: &VM) -> Result<Val, Box<VMError>> {
-        Ok(Double::new(vm, self.val.sqrt()))
+        if self.val < 0.0 {
```
ltratt: Is this branch necessary? I think `self.val.sqrt()` returns `NaN` if `self.val < 0` anyway?

src/lib/vm/objects/integers.rs
```diff
@@ -156,7 +156,11 @@ impl Obj for ArbInt {
     }
 
     fn sqrt(&self, vm: &VM) -> Result<Val, Box<VMError>> {
-        ArbInt::new(vm, self.val.sqrt())
+        if self.val < Zero::zero() {
```
ltratt: Same comment as above.

src/lib/vm/val.rs
```diff
@@ -424,7 +424,11 @@ impl Val {
     /// Produces a new `Val` which is the square root of this.
     pub fn sqrt(&self, vm: &VM) -> Result<Val, Box<VMError>> {
         if let Some(lhs) = self.as_isize(vm) {
-            return Val::from_isize(vm, lhs.sqrt());
+            if lhs < 0 {
```
ltratt: Same comment as above.

src/lib/vm/objects/integers.rs
ummarikar: Rust throws a 'number is imaginary' error without this if statement

ummarikar:
Removing all the above if statements produced the following error:
`thread 'main' panicked at 'the square root of a negative is imaginary', /Users/umarmarikar/.cargo/registry/src/github.com-1ecc6299db9ec823/num-integer-0.1.41/src/roots.rs:171:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.`
However, all the negative test for doubles worked.

src/lib/vm/objects/integers.rs
ltratt: OK.

ltratt:
Let's remove the `if`s that aren't needed, but we should definitely keep those that are!

ummarikar:
Done!

ummarikar:
Squashed

ltratt:
I just realised a problem. If `self.val < 0` (for ints or big ints), we need to throw an error, not return a `NaN`. It's kinda weird if `-1.sqrt()` returns a double. We'll need tests for that too. We can make a new error (borrowing shamelessly from Python) `VMError::DomainError` or similar.

ltratt:
If this is ready for rereview, you need to tell me explicitly.

ummarikar:
> If this is ready for rereview, you need to tell me explicitly.

I know, I only just got round to doing a double check now (I've had coursework from my modules recently), and I think there's nothing more I need to do so feel free to take a look!

src/lib/vm/core.rs
```diff
@@ -41,6 +41,8 @@ pub enum VMError {
     /// A value which can't be represented in an `usize`.
     CantRepresentAsUsize,
     DivisionByZero,
+    /// A value which is mathematically undefined
```
ltratt: Missing `.` at the end of the sentence.

smarr:
I saw this before, and thought it is sensible.

However, triggered by https://github.com/SOM-st/SOM/issues/24, does yksom give the expected results for `4 sqrt`, `1 sqrt`, and `3 sqrt`?

```
---> 4 sqrt
it = 2
---> 1 sqrt
it = 1
---> 3 sqrt
it = 1.7320508075688772
```

and:

```
---> 4 sqrt class
it = Integer
---> 1 sqrt class
it = Integer
---> 3 sqrt class
it = Double
```

ltratt:
Good point! No, this PR will not produce ints even when it could. I'm not even entirely sure I know the best way to implement this semantically. Obvious possibilities are 1) see if the FP number produced is int-like and if so convert it 2) convert the result to an int, square it, and see if you get the original answer again. I *think* those two routes will behave subtly different in some edge conditions, though I haven't checked. Any thoughts?

smarr:
SOM and CSOM do basically the same for the case with an Integer as a receiver.

https://github.com/SOM-st/CSOM/blob/master/src/primitives/Integer.c#L267-L274

https://github.com/SOM-st/som-java/blob/3fab8ac7f34cedbbe15719fc607bceef25343d59/src/som/vmobjects/SInteger.java#L82-L89

Note that `4.0 sqrt class` yields a `Double`.

smarr:
Ah, and at this point, in the sense of a spec, I wouldn't consider this required behavior...

I think, I'd need to find the people who implemented that originally and get their reasoning, i.e., do some historical digging.

ltratt:
> https://github.com/SOM-st/CSOM/blob/master/src/primitives/Integer.c#L267-L274

OK, so that's option #1 of the two I suggested. It's a reasonable enough definition, at least from this PR's perspective. @ummarikar hopefully that gives you a bit of feedback for how to improve the PR a bit?

> Ah, and at this point, in the sense of a spec, I wouldn't consider this required behavior...

Fair enough. I think it's not a bad idea to implement it similarly though -- if nothing else, reducing the points of divergence (as far as possible) to only clear bugs will probably make everybody's life easier!

ummarikar:
> > https://github.com/SOM-st/CSOM/blob/master/src/primitives/Integer.c#L267-L274
> 
> OK, so that's option #1 of the two I suggested. It's a reasonable enough definition, at least from this PR's perspective. @ummarikar hopefully that gives you a bit of feedback for how to improve the PR a bit?
> 
> > Ah, and at this point, in the sense of a spec, I wouldn't consider this required behavior...
> 
> Fair enough. I think it's not a bad idea to implement it similarly though -- if nothing else, reducing the points of divergence (as far as possible) to only clear bugs will probably make everybody's life easier!

Yes, this is great I'll implement it asap. Thanks for the feedback!

ummarikar:
Just did some playing around with yksom. It has the following behaviour:
`---> 4 sqrt class
it = Integer
---> 1 sqrt class
it = Integer
---> 3 sqrt class
it = Integer`
I'll make sure that Integer sqrt produces a Double if the result is not a whole number.

ltratt:
> I'll make sure that Integer sqrt produces a Double if the result is not a whole number.

The links Stefan sent to other SOM's might give you a good pointer for how to go about this.

ummarikar:
Added some code to val.rs which ensures that a Double is returned when the square root of an Integer is not a whole number. I was not sure if you want me to change the implementation of sqrt for BigInts when the result is not a whole number. The current implementation in Java SOM is to return an error in this case.

ummarikar:
I also added an extra test case for "3 sqrt".

ltratt:
We need to make sure both `Int`s and `ArbInt`s have the same behaviour (from the SOM user's perspective they're one and the same).

ltratt:
I think we also need checks along the lines of:

```
16 sqrt class println.
```

to make sure that things are printing `Integer` / `Double` as appropriate.

to check that

ummarikar:
Sure I'll start that now.

lang_tests/int29.som
```diff
@@ -0,0 +1,18 @@
...
+    Double
+    Double
+"
+
+int29 = (
+    run = (
+        (4 sqrt class) println.
+        ((1 << 200) sqrt class) println.
+        (3 sqrt class) println.
+        ((1 << 103) sqrt class) println.
```
ummarikar: This will fail with a bit shift to the left of 105 (it returns an Integer). I am unsure of what I need to do to solve this and would appreciate some guidance.

ltratt: I doubt there's going to be a perfect answer here, because we don't have `BigDouble` (or any equivalent). I suspect that the least worst answer *might* be just to say that `ArbInt::sqrt` always returns an `ArbInt`. That will certainly give mildly surprising (i.e. integer) results for "small" `ArbInt`s, but it will behave predictably even as an `ArbInt` becomes bigger.

ummarikar: There is the BigDecimal in Rust (https://docs.diesel.rs/bigdecimal/index.html). Is this something I should look into? Or, should I change the code so that `ArbInt::sqrt` returns an `ArbInt` in all cases?

ltratt: SOM doubles are IEEE 754 (i.e. fixed size), so we don't want to go down the `BigDecimal` route. I think it's OK if `ArbInt::sqrt` returns `ArbInt` in all cases.

ummarikar: Understood! Will change this now.

ummarikar:
All done.

ltratt:
Please squash.

ummarikar:
Squashed.

PR #58, opened 2019-10-27
Statically calculate stack sizes

ltratt:
This PR is mostly about precalculating the maximum stack size a block can possibly use at run-time so that we can remove some dynamic checks. This allows to use a fixed-size SOM stack, which means that we can avoid checking whether resizing (etc.) is necessary
(i.e. we can push elements onto the stack without any branching). This gives us roughly a 2.5% gain on our little while benchmark.

The main bulk of this PR is in https://github.com/softdevteam/yksom/commit/7d4ab33a6b5da7012d88788c9d9224398392f14c; the other commits are relatively minor.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ regex = "1.1"
 [dependencies]
 abgc = { git="https://github.com/softdevteam/abgc" }
 abgc_derive = { git="https://github.com/softdevteam/abgc" }
+arrayvec = "0.5"
 cfgrammar = "0.4"
 getopts = "0.2"
 indexmap = "1.0"
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -8,6 +8,7 @@
 // terms.
 
 use std::{
+    cmp::max,
     collections::hash_map::{self, HashMap},
     path::Path,
 };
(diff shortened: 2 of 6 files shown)


smarr:
Could you document the calculation? I didn't find the summarized logic.

SOMs traditionally do this, though, they need to leave head room for `#escapedBlock`, `#doesNotUnderstand`, and similar errors:

 https://github.com/SOM-st/som-java/blob/master/src/som/vm/Universe.java#L472-L477

The "fudge" factor is currently `2`.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ regex = "1.1"
 [dependencies]
```
ptersilie: Typos in commit message: `have to have`

ltratt:
> Could you document the calculation?

It's all contained in `c_expr` (https://github.com/softdevteam/yksom/pull/58/files#diff-1496b1d005cb52b52e1e8b08610a834eR338). It's a relatively simple calculation although, inevitably, checking things again this morning showed at least one mistake that I'd made (and one way that I could do the calculation more accurately). @ptersilie That new commit will need squashing away.

> The "fudge" factor is currently 2.

Aha, that will be very useful to know when I get to those primitives -- thanks!

Cargo.toml
ltratt: Oops!

ptersilie:
I don't think I understand b259f6b.

ltratt:
Which part?

ptersilie:
All of it. Your commit message talks about a fix and a change, but I have no idea which one is which and what they are supposed to fix/change.

ltratt:
Point taken. `KeywordMsg`'s calculation was broken because I forgot to take into account the cost of evaluating the receiver *and* the cost of keeping the receiver on the stack. `BinaryMsg` was suboptimal (i.e. also incorrect, but not in such an obviously bad way) because I added together the LHS and RHS stack sizes, forgetting that they're evaluated one at a time.

ptersilie:
Aha, just one more question. Why do we need to add `1` as well now?

ltratt:
Which addition of 1? [There are 2.]

ptersilie:
Both.

ltratt:
For `BinaryMsg`, because you have the LHS on the stack when evaluating the RHS.

For `KeywordMsg`, because you have the receiver on the stack while evaluating arguments.

ptersilie:
Okay, think I got it now. Please squash.

ltratt:
Squashed.

PR #59, opened 2019-10-30
Reading from an unassigned SOM variable should return 'nil'.

ltratt:
Previously we threw an error in such cases. While, arguably, that might be better language design, it doesn't match SOM's semantics. So this commit fixes that.

The implementation is not particularly efficient in a reference counted GC (we increment nil's reference count only to decrement it on the first assignment to a var), but it's probably not worth optimising this at this stage.

### DIFF
--- a/lang_tests/unassigned_var.som
+++ b/lang_tests/unassigned_var.som
@@ -1,10 +1,9 @@
 "
 VM:
-  status: error
-  stderr: UnassignedVar(1)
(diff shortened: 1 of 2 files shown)


smarr:
SOM follows Smalltalk here.
"Everything" is implicitly initialized with `nil`.
"Everything" includes at least variables, fields, and array elements.

PR #60, opened 2019-11-03
Fix class names.

ltratt:
SOM class names should match the file name.

### DIFF
--- a/lang_tests/block10.som
+++ b/lang_tests/block10.som
@@ -4,7 +4,7 @@ VM:
     7
 "
 
-block8 = (
+block10 = (
     f = (
         self g: [ :x :y | ^ (x - y) ].
         ^'b'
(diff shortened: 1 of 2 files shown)


smarr:
Risking just being annoying:

**Class names should be capitalized** in any Smalltalk, and SOM.
Also, numbering classes is likely to mean that the code is less maintainable, because noone remembers what the difference between the classes were intended to be.

ltratt:
The comment is a fair one. I have a deep personal distaste for capital letters in file names verging on the pathological -- I can't help it! I can't bring myself to change the file names, but if someone else fixes them, I will merge it happily.

The "numbers as part of names" thing is true, but it's often *really* hard to think of meaningful names for so many tests. I agree that we should probably try a bit harder than we do, but I'm realistic that we'd always end up with *some* numbers in names.

ptersilie:
> I have a deep personal distaste for capital letters in file names verging on the pathological

I hope you'll never have to write Java then. ;-)

PR #61, opened 2019-11-03
Simplify bytecode

ltratt:
This PR slightly simplifies yksom's bytecode (well, not that it's "byte"code, but you get the idea) by getting rid of the idea of a constant pool of integers and strings. This seemed like a good idea when I didn't understand SOM, but it's now clearly overkill: the idea of pooled integers, in particular, makes very little sense in the presence of tagged integers. Once integers are removed, then the only thing left is strings, at which point we don't have a constant pool but a string pool. The two commits make these two changes in order.

### DIFF
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -30,9 +30,9 @@ pub struct Compiler<'a> {
     /// We collect all message sends together so that repeated calls of e.g. "println" take up
     /// constant space, no matter how many calls there are.
     sends: IndexMap<(String, usize), usize>,
-    /// We map constants to an offset so that we only store constants once, no matter how many
-    /// times they are reference in source code.
-    consts: IndexMap<cobjects::Const, usize>,
+    /// We map strings to an offset so that we only store them once, no matter how many times they
+    /// are referenced in source code.
+    strings: IndexMap<String, usize>,
     /// All the blocks a class contains.
     blocks: Vec<cobjects::Block>,
     /// The stack of variables at the current point of evaluation.
@@ -54,7 +54,7 @@ impl<'a> Compiler<'a> {
             path,
             instrs: Vec::new(),
             sends: IndexMap::new(),
-            consts: IndexMap::new(),
+            strings: IndexMap::new(),
             blocks: Vec::new(),
             vars_stack: Vec::new(),
             closure_depth: 0,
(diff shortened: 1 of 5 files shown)


smarr:
In classic SOM implementations, methods have a number of "literals" associate with them.
Depending on the implementation, those are `Integer`, `Double`, `String`, or `Symbol`  (interned strings) objects. If I am not missing anything, those are the types of things that can appear literally in SOM code.

The `PUSH_CONSTANT` bytecode would then read those from the method via an index into the literal array (https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Interpreter.java#L89)

ltratt:
Yes -- but I claim it is easier to understand if `Double`, `Int`, (and maybe `Symbol`!) are handed as separate bytecodes. It's purely a code readability thing: I managed to confuse myself with how the constant pool was implemented, and what was in there, and I implemented it -- which was a bit embarrassing!

smarr:
If you want to type things, sure, type things.
If you want to make things understandable, help people to understand it, for instance with useful debug output.
Then the difference between one or multiple bytecodes becomes minor, and the benefit of having only a single implementation of one concept means there's less code.

I don't know whether one or the other option is easier to understand, less code vs. more explicit code can make a difference, I suppose.

```
$ ./som.sh -d -d -cp core-lib/Smalltalk Examples/Hello.som

...
45:PUSH_CONSTANT     (index: 8) value: (#String) "Unable to resolve "
...
12:PUSH_CONSTANT     (index: 6) value: (#Integer) 3
...
38:PUSH_CONSTANT     (index: 10) value: (#Symbol) #run:
```
```

**Comment by ltratt** [comment]
> I don't know whether one or the other option is easier to understand, less code vs. more explicit code can make a difference, I suppose.

Agreed. This could be one of those things where I change my mind in a few weeks time! At the moment it feels like an improvement though.

File: src/lib/compiler/ast_to_instrs.rs [hunk 1]
```diff
@@ -30,9 +30,9 @@ pub struct Compiler<'a> {
     /// We collect all message sends together so that repeated calls of e.g. "println" take up
     /// constant space, no matter how many calls there are.
     sends: IndexMap<(String, usize), usize>,
-    /// We map constants to an offset so that we only store constants once, no matter how many
-    /// times they are reference in source code.
-    consts: IndexMap<cobjects::Const, usize>,
+    /// We map strings to an offset so that we only store themonce, no matter how many times they
```
ptersilie: > themonce

Typo.

> clearer to

Also another typo in the commit message.

ptersilie:
Just some typos.

ptersilie:
Okay, please squash (and fix the other typo during that).

ltratt:
I've fixed the code comment typo. I can fix the commit message in squashing if you're OK with that?

ltratt:
Squashed.

PR #62, opened 2019-11-04
Add an inline cache

ltratt:
This PR adds simple support for inline caches during method lookup. The main commit is https://github.com/softdevteam/yksom/commit/0950f3de11a4f8dd75476e166df84a1a1e46cece and it's probably worth starting with that commit's message and then going backwards to the two preceding commits (which are really just warm-ups for that main commit). This is an important optimisation: on our simple while loop, it improves performance by over 20%.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -261,20 +261,20 @@ impl VM {
     fn exec_user(&self, rcv: Val, cls: &Class, meth_start_pc: usize) -> SendReturn {
         let mut pc = meth_start_pc;
         let stack_start = self.stack_len();
-        while let Some(instr) = cls.instrs.get(pc) {
-            match *instr {
-                Instr::Block(blkinfo_off) => {
-                    let blkinfo = cls.blockinfo(blkinfo_off);
+        while let Some(ref instr) = cls.instrs.get(pc) {
+            match instr {
+                &Instr::Block(blkinfo_off) => {
+                    let blkinfo = cls.blockinfo(*blkinfo_off);
                     self.stack_push(Block::new(
                         self,
                         Val::recover(cls),
-                        blkinfo_off,
+                        *blkinfo_off,
                         Gc::clone(&self.current_frame().closure),
                         blkinfo.num_params,
                     ));
                     pc = blkinfo.bytecode_end;
                 }
-                Instr::Builtin(b) => {
+                &Instr::Builtin(b) => {
                     self.stack_push(match b {
                         Builtin::Nil => self.nil.clone(),
                         Builtin::False => self.false_.clone(),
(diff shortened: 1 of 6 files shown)


src/lib/vm/core.rs
```diff
@@ -303,38 +303,53 @@ impl VM {
...
-                    let (meth_cls_val, meth) =
-                        stry!(stry!(cls.downcast::<Class>(self)).get_method(self, &name));
+                    let rcv_cls = rcv.get_class(self);
+                    // Implement our simple inline cache which just remembers the last class used
+                    // at this particular message send: if the cache is empty, or the receiver
+                    // class we find doesn't match, we update the cache.
+                    let (meth_cls_val, meth) = loop {
+                        let cache_cell = unsafe { &mut *cache.get() };
+                        if let Some((cls, (meth_cls_val, meth))) = cache_cell {
+                            if cls.bit_eq(&rcv_cls) {
```
smarr: Following from the comment, that's a pointer comparison, I guess.

src/lib/vm/core.rs
```diff
@@ -303,38 +303,53 @@ impl VM {
...
+                                break (meth_cls_val.clone(), Gc::clone(meth));
```
smarr: Are these clone operations copying ref-counted pointers?

src/lib/vm/core.rs
```diff
@@ -328,13 +328,28 @@ impl VM {
...
                     let rcv = self.stack_pop_n(*nargs);
 
-                    let cls = rcv.get_class(self);
-                    let (meth_cls_val, meth) =
-                        stry!(stry!(cls.downcast::<Class>(self)).get_method(self, &name));
+                    let rcv_cls = rcv.get_class(self);
+                    // Implement our simple inline cache which just remembers the last class used
+                    // at this particular message send: if the cache is empty, or the receiver
+                    // class we find doesn't match, we update the cache.
+                    let (meth_cls_val, meth) = loop {
```
ptersilie: Why are we looping here? Due to the breaks it doesn't appear as if we can ever have more than one iteration.

ltratt: I wondered if you'd spot this :) This is just to reduce code duplication because Rust has no `goto`. It's a little difficult to explain, but see if you can find a way of rewriting the code without having to write the "[re]fill the cache" part twice. If you can, and it doesn't involve `loop`, I would like to know!

src/lib/vm/core.rs
ltratt: It's sort-of a pointer comparison: since it's a `Val` and that is a tagged reference, it might also be a non-pointer comparison. But, if the two things involves are pointers, then yes it's a pointer comparison. [If anyone has suggestions for a better name for this function, I'm all ears! `bit_eq` is a little generic sounding.]

src/lib/vm/core.rs
ltratt: Yes.

src/lib/vm/core.rs
ptersilie: I already wondered if that was the reason but thought I'd better ask in case I missed something. I don't see a better way of writing this either.

ltratt: What I should have said explicitly that `loop` and `break` are the closest things Rust has to `goto`. This is the first time I've felt that this trick is useful: hopefully I won't use it very often :)

PR #64, opened 2019-11-06
Add && operator for Boolean objects.

ummarikar:
Add && for Boolean objects along with a test for all possible true and false combinations.

### DIFF
--- a/lang_tests/bool4.som
+++ b/lang_tests/bool4.som
@@ -1,0 +1,33 @@
+"
+VM:
+  stdout:
+    0
+    1
+    true
+    false
+    false
+    false
+    true
+    false
+    false
+    false
+"
+
+bool4 = (
+    run = (
+        | x |
+        x := 0.
+        false && [ x := 1 ].
+        x println.
+        true && [ x := 1 ].
+        x println.
+        (true && [ true ]) println.
+        (true && [ false ]) println.
+        (false && [ true ]) println.
+        (false && [ false ]) println.
+        (true && true) println.
+        (true && false) println.
+        (false && true) println.
+        (false && false) println.
+    )
+)
--- a/lib/SOM/Boolean.som
+++ b/lib/SOM/Boolean.som
@@ -3,4 +3,6 @@ Boolean = (
         self ifTrue: [ ^trueBlock value  ].
         self ifFalse: [ ^falseBlock value ].
     )
+
+    && boolean = ( ^self and: boolean )
 )


lang_tests/bool4.som
```diff
@@ -0,0 +1,17 @@
...
+    false
+    false
+"
+
+bool4 = (
+    run = (
+        (true && [true]) println.
+        (true && [false]) println.
+        (false && [true]) println.
+        (false && [false]) println.
```
ummarikar: In SOM, if I just write ```(true && true) println.``` for example, this works fine. I tried this in yksom and it did not work. I got the following error: ```UnknownMethod("value")```. It only worked once I put the square brackets around the argument. I am not sure how to fix this behaviour.

ltratt: I can see the problem. Fix coming shortly.

smarr: `Object` should implement a `value` method returning `self`: https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L48

smarr: I'd recommend to keep the block tests, because operators like `&&` are the first ones where people start to cheat. Checking that blocks are supported explicitly is a good thing here.

ltratt: Good point. We can have those as well as the non-block tests.

smarr: Thinking about it, it may also be useful to test for the short-circuit semantics.
With that, I mean that the block will only be evaluated when necessary.
For instance, `false && [self error]` won't trigger the error, because the block is not executed.

ltratt: I agree that would be useful, but I don't think yksom has the necessary machinery to test this yet unfortunately :/

smarr: Ehm, you do:
https://github.com/softdevteam/yksom/blob/master/lib/SOM/False.som#L5
There's no magic, just polymorphic methods.

ltratt: Ah, hang on, I was too quick: I thought you meant we needed `self error`. I think what you're saying is that we want a test along the lines of the following:

```
x := 0.
false && [x := 1].
x println. "Should print 0"
true && [ x := 1].
x println. "Should print 1"
```

?

smarr: Ah, well, do you have local variables?

```Smalltalk
ShortCircuit = (
    run = (
        | var |
        var := 1.
        (false && [ var := 2 ]) println.
        var println.
    )
)
```

ltratt: Great minds, great minds! Yes, we have local variables, so such a test should work. @ummarikar I agree with Stefan that tests along these lines would be good additions to the PR.

smarr: Yepp, that's all a test would take, I think.

ummarikar: Thanks, guys! I'll get started on this.

ltratt:
@ummarikar Not to worry about this time, but you mustn't force push to a branch unless a reviewer explicitly tells you to. The reason for that is that it's then not possible for the reviewer to see what's changed. It's fine to keep adding changes as new commits: at the end of the PR you'll be given the opportunity to squash them down to a smaller set (if that's the right thing to do).

ummarikar:
@ltratt apologies for the force push. I incorrectly assumed it would be fine since all I was doing was rebasing the commit in this pull request on top of softdevteam/yksom's current master branch (in order to get the changes made in #65). Rest assured I did not make any further changes and will seek your approval in the future.

ummarikar:
Updated the tests!

ltratt:
Not a problem in this instance. The trick is to just do `git merge` and leave the merge commit in. When you eventually squash/rebase it the end of the PR, gith will remove those merge commits automatically (unless you explicitly ask it to keep them).

ltratt:
These look fine, though we should follow the SOM convention of putting spaces inside square brackets i.e. `[ blah ]` (not `[blah]`). Once that's done, I think we'll be OK for this PR.

ummarikar:
Yes, of course, I did not think of that as I was too busy trying to avoid a merge commit! I'll remember this for next time, thanks.

ummarikar:
> These look fine, though we should follow the SOM convention of putting spaces inside square brackets i.e. `[ blah ]` (not `[blah]`). Once that's done, I think we'll be OK for this PR.

Will change this now.

ummarikar:
All done.

ltratt:
Please squash.

ummarikar:
Squashed.

ltratt:
The commit message needs a bit of trimming: it's very repetitive. Please force push an update to it.

ummarikar:
Yes I just noticed it my mistake I forgot to shorten it

ummarikar:
All done.

PR #65, opened 2019-11-06
Objects evaluate to themselves unless overridden.

ltratt:
@ummarikar You can review this one. Check it over carefully: if you are happy with it then you can say bors r plus (with a '+' not 'plus'!).

bors delegate @ummarikar


ltratt:
BTW, this should fix the problem in https://github.com/softdevteam/yksom/pull/64

ltratt:
bors delegate=ummarikar

PR #66, opened 2019-11-06
Add placeholders for all the unimplemented primitives.

ltratt:
This stops us getting endless "unknown primitives" when trying to import the full SOM library. Actually trying to make use of any of these primitives will lead to a Rust-level "unimplemented" panic.

With this the only remaining compile-time error from the full SOM library is:

```
File 'lib/SOM/Object.som', line 44, column 21:
  , element = ( ^(Vector new append: self) append: element )
Unknown variable 'Vector'
```

@ummarikar This is another you can review. If you spot *anything* you are not sure about (something wrong / missing / etc.) shout. I am not perfect, and regularly make mistakes!

bors delegate=ummarikar


ltratt:
OK, there's a lie in the above: this stops us getting warned about primitives in the classes we currently load in. There are classes we currently don't load in that have additional primitives, but we can deal with those later.

ummarikar:
It looks good to me. I am feeling like I am gradually getting the hang of Rust and this project which is exciting!

PR #68, opened 2019-11-08
Various suggestions from Clippy.

ltratt:
Note that Clippy has a pronounced dislike of comparing floats: in our case, such comparisons are entirely sensible. Turning this warning off is the only way to maintain our sanity.


ltratt:
Because Clippy noticed one pointless `clone` in a critical path, this PR even speeds up our while loop benchmark by ~1%! I think that's a first...

PR #70, opened 2019-11-11
Sqrt should return an Int where possible.

ummarikar:
Sqrt should return an Int if the sqrt of a BigInt can be represented as an isize.


ltratt:
Good spot!

bors r+

PR #71, opened 2019-11-12
Faster tagged pointers

ltratt:
This PR changes the integer/pointer tagging scheme: previously pointers were tagged with a `0` and integers with a `1`. This PR inverts that relationship (https://github.com/softdevteam/yksom/commit/5aee9ffa887dd9ee5876c4dd8129a6b15f27f091), which then allows us to use normal integer operations on tagged integers (https://github.com/softdevteam/yksom/commit/7d7dfdbd3d9df87da292d26d65e36990efa385ca). [The other commits are either setting scene for the commits or fixing minor issues.]

I've long been dimly aware of this scheme, though I've no idea where I first heard about it. There seems to be very little about it lurking around: the best description I've found is [this LtU post](http://lambda-the-ultimate.org/node/3201) (though note that you need to read the comments to correct one or two minor mistakes).


ptersilie:
Do you have any data on how this changes the performance. While intuitively I can see why this should improve performance, some comments in the link provided suggest it's only better in the average case, but not in highly optimised cases (though those comments didn't do a very good job of explaining why so I'm not sure if that's true).

src/lib/vm/val.rs
```diff
@@ -81,6 +81,14 @@ impl Val {
         }
     }
 
+    /// If this `Val` is a `GCBox` then convert it into ` ThinObj`. The caller of this function
```
ptersilie: Two typos here: an extra space before `ThinObj`; and `GXBox`.

ltratt:
The commit message for https://github.com/softdevteam/yksom/pull/71/commits/7d7dfdbd3d9df87da292d26d65e36990efa385ca has some info -- on our simple while benchmark, performance improves by 2.5% or so.

src/lib/vm/val.rs
ltratt: Oops! Fixed in https://github.com/softdevteam/yksom/pull/71/commits/ddd2d7b31135277eef9411ffc7e21b33fb5b8e0c.

ptersilie:
> The commit message for [7d7dfdb](https://github.com/softdevteam/yksom/commit/7d7dfdbd3d9df87da292d26d65e36990efa385ca) has some info -- on our simple while benchmark, performance improves by 2.5% or so.

Thanks. I noticed this also after I posted my message.

ptersilie:
Looks good. Please squash!

ltratt:
Squashed.

PR #73, opened 2019-11-13
Add bitwise and operator.

ummarikar:
Add bitwise and operator - add tests for Ints and BigInts and tests for when the argument is not an Int.


src/lib/vm/objects/mod.rs
```diff
@@ -90,6 +90,11 @@ pub trait Obj: std::fmt::Debug + abgc::GcLayout {
         unimplemented!();
     }
 
+    /// Produce a new `Val` which performs a bitwise and with `other` and this
```
ltratt: Missing "." at end of sentence.

ummarikar:
Done (apologies for missing that!).

ltratt:
Quick question: have you checked this against another SOM (e.g. Java SOM)?

ummarikar:
> Quick question: have you checked this against another SOM (e.g. Java SOM)?

Yes indeed. All features I implement I check them against Java SOM before committing the changes. I run against Java SOM, copy the output into my tests, then implement the feature to get the tests to pass. Is that the right way to go about it?

ltratt:
That's the right way to do it IMHO (though you should keep a questioning mind -- Java SOM isn't always correct itself!).

Please squash.

ummarikar:
Understood, I will definitely keep this in mind.

ummarikar:
Squashed.

PR #74, opened 2019-11-14
Reorder functions in structs

ltratt:
After renaming `bit_xor` to `xor`, this PR simply sorts out the order that functions are written in so that we're more consistent across structs -- this PR has no functional change. @ummarikar This one is for you to review.

bors delegate=ummarikar


ummarikar:
This looks good. Do you need to squash?

ltratt:
No squashing needed in this case thanks!

PR #75, opened 2019-11-15
Remove one layer in the compilation process

ltratt:
Previously the yksom compiler produced `cobject`'s, a post-AST, but pre-run-time data structure that the interpreter than converted into its internal format and ran. However, this hasn't turned out to be a particularly useful layer: `cobject`'s are almost identical in structure to run-time `object`'s (3 out of the 4 `cobject`'s are literally identical). This PR thus removes this intermediate layer entirely: the compiler now creates run-time values directly.

There is probably a tiny performance benefit from doing this, but it's almost certainly too small to measure. The advantage of this PR is mostly code simplification (100 lines fewer code overall!). However, it also makes one future thing easier: SOM dynamically loads classes and previously this felt slightly awkward due to the `cobject` layer. By removing that layer, it should be more obvious how to do dynamic class loading in the future.


ltratt:
@ummarikar You might want to have a look at this one and offer comments (though @ptersilie is the main reviewer).

ltratt:
bors try

ummarikar:
With my current understanding, everything seems fine. But, I will need to go deeper into the workings of yksom to provide value in future PRs!

src/lib/compiler/cobjects.rs
```diff
@@ -11,24 +11,16 @@ use std::path::PathBuf;
 
 use crate::compiler::instrs::Instr;
 
-pub use crate::vm::objects::{Method, MethodBody};
+pub use crate::vm::objects::{BlockInfo, Method, MethodBody};
```
ptersilie: The commit message says `vm::objects::class::BlockInfo` but this imports ``vm::objects::BlockInfo``. I presume the latter is correct?

ltratt: They both point to the same thing underneath (the latter is shorthand for the former).

ptersilie: Nevermind, just saw that `objects` imports `BlockInfo`.

ptersilie:
Looks fine.

bors r+

ptersilie:
It seems nothing has changed since the last commit, yet the build still fails. Odd.

ltratt:
It's because #73 was merged after this PR was raised and that PR references the old layout. I need to rebase against master to fix this (it's a one line fix but I can't push it without force pushing). OK?

ptersilie:
Go ahead!

ltratt:
I meant https://github.com/softdevteam/yksom/pull/76 (which was raised after this PR, but was merged first).

ltratt:
It's hard to see but one line (237) in https://github.com/softdevteam/yksom/pull/75/commits/dabe52e6ac10733909eaedd63d628f65afc4dc05 has now changed. I think this is ready to try again.

PR #76, opened 2019-11-17
Add mod operator for Integers and Doubles in yksom.

ummarikar:
Add mod operator for Integers and Doubles and include tests for all possible combinations of BigIntegers, Doubles and Integers. I have named the method for mod in yksom `modulus` as `mod` is a keyword in Rust.


src/lib/vm/objects/integers.rs
```diff
@@ -111,6 +111,23 @@ impl Obj for ArbInt {
...
     }
 
+    fn modulus(&self, vm: &VM, other: Val) -> Result<Val, Box<VMError>> {
+        if let Some(rhs) = other.as_isize(vm) {
+            ArbInt::new(vm, &self.val % rhs)
+        } else if let Some(rhs) = other.try_downcast::<ArbInt>(vm) {
+            ArbInt::new(vm, &self.val % &rhs.val)
+        } else if let Some(rhs) = other.try_downcast::<Double>(vm) {
+            match self.val.to_f64() {
+                Some(i) => Ok(Double::new(vm, i % rhs.double())),
```
ltratt: I *think* this is incorrect: `Int % Double` should be an int. I think this is the same bug as https://github.com/softdevteam/yksom/pull/71/commits/bddaa3c2ba277361a5bbed0bc9482d16516629d5.

ltratt:
From this PR on (bad luck!), I think we should follow @smarr's suggestion and give tests more descriptive names. So it might be `double_modulus.som` and so on rather than `double9283.som`.

src/lib/vm/objects/integers.rs
ummarikar: From what I could tell in the `bddaa3c` fix, the behaviour was different between BigInts and Ints which shouldn't be the case for division operators. But, in this PR `Int % Double` is a Double as well as `BigInt % Double` (this is the same behaviour as Java SOM). However, I am more than happy to change it if you believe its the best option!

ltratt: Ah, that's a good point. @smarr do you have a preference here, or is Java SOM's behaviour correct as-is?

ltratt:
Don't we also need `double_modulus_err.som` and `arbint_modulus_err.som` tests?

src/lib/vm/objects/integers.rs
smarr: I am not sure why you would `Int % Double` to result in an `Int`.
What should `5 % 4.5` be? I'd guess `0.5`?

Pharo seems to do:

```
Int % Double => Double
Double % Int => Double
```

ummarikar:
> Don't we also need `double_modulus_err.som` and `arbint_modulus_err.som` tests?

We do indeed, I will do that now.

src/lib/vm/objects/integers.rs
ltratt: OK, let's stick with that then. Thanks for the detective work!

ummarikar:
@ltratt Just finished adding the new tests. What is your preference for class names as well? I have left them numbered but I can change them to something more appropriate. So, for example, something like the following for double_modulus.som:
```
double_modulus = (
    run = (
        ...
    )
)
```

ltratt:
In SOM the class name needs to match the file name. [yksom doesn't enforce this -- yet -- but other SOMs do, so we might as well get it right.]

ummarikar:
Sounds good to me. Commit coming shortly...

ummarikar:
Do we also want to make our classes and filenames camel case with capital letters as well to match Java SOM? I can start that from now if so.

ltratt:
I think capitalisation only makes sense if we do it all in one go. Let's worry about that later.

ummarikar:
Done.

ltratt:
Please squash.

ummarikar:
Squashed.

PR #77, opened 2019-11-18
Make method execution only require method pointers.

ltratt:
Previously, in order to execute a method you had to carry around its class, since the method's program counter was relative to the class -- in other words one would use a `(Class, Method)` pair to execute a method. The original idea for this was because this makes it easier to unload classes -- which SOM never does. So it's extra complexity, and an extra performance overhead, for no gain.

This PR thus simplifies the VM so that methods can be executed without needing a pointer to the class. This has a number of knock-on effects because various things that modules need (e.g. sends, strings) were also stored on classes. Most of the commits in this PR thus move one thing at a time from classes to the main VM. In a sense, the easy bit is in the final commit (b9a3aa61), and it might be easiest to start there and then move to the first commit.

Overall this PR gives a 10% speedup on our simple while benchmark. Most of this is probably largely because it reduces the amount of reference counting increments/decrements we perform.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,6 @@ abgc_derive = { git="https://github.com/softdevteam/abgc" }
 arrayvec = "0.5"
 cfgrammar = "0.4"
 getopts = "0.2"
-indexmap = "1.0"
 itertools = "0.8"
 lrlex = "0.4"
 lrpar = "0.4"
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -14,7 +14,6 @@ use std::{
 };
 
 use abgc::Gc;
-use indexmap::{self, map::IndexMap};
 use itertools::Itertools;
 use lrpar::{Lexeme, Lexer};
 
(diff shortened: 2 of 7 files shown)


ptersilie:
Looks good.

bors r+

smarr:
Wait, what? Could you please explain where the program counter is?
Or asked differently: do you support recursive method/block activations?

ltratt:
Previously the program counter was a `(Class, usize)` pair. The program counter is now, literally, a `usize`: there's a global list of instructions. So there're no implications on recursion.

> Or asked differently: do you support recursive method/block activations?

Yes, https://github.com/softdevteam/yksom/blob/master/lang_tests/block6.som and https://github.com/softdevteam/yksom/blob/master/lang_tests/block7.som show that. Or, if they don't, let me know what you're thinking of, and I'll happily write a new test to check it does work! This wouldn't be the first time I've made a major mistake!

smarr:
I was more thinking along the lines of (untested):

```Smalltalk
Fibonacci = (
  start = (
    | fibBlock |
    fibBlock := [:n | 
      n <= 1
            ifTrue:  1
            ifFalse: [ (fibBlock value: n - 1) + (fibBlock value: n - 2) ] ].
    ^ fibBlock value: 8
  )
)
```

PR #78, opened 2019-11-18
Add Fibonacci test.

ltratt:
As suggested by @smarr in https://github.com/softdevteam/yksom/pull/77#issuecomment-555062043. The test passes fine.

### DIFF
--- a/lang_tests/fib.som
+++ b/lang_tests/fib.som
@@ -1,0 +1,27 @@
+"
+VM:
+  status: success
+  stdout:
+    1
+    2
+    8
+    34
+"
+
+fib = (
+    run = (
+        (self fib: 1) println.
+        (self fib: 2) println.
+        (self fib: 5) println.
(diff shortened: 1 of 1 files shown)


ltratt:
bors try

ltratt:
@smarr Hopefully the `fib` test succeeding is a decent indication that this part works OK? I must admit, I wasn't even sure if yksom had all the bits it needed for such a test, but it turns out that it does. I should have added it ages ago!

ptersilie:
Looks good. If @smarr is happy, I will merge.

smarr:
Yeah, I was just concerned about how many program counters there can be for a block/method.
There need to be `n` program counters, where `n` is the number of activations of the block/method, which I suppose is checked by the test. If it wouldn't, I'd expect an infinite loop, I think.

In terms of language features, I don't think it uses anything I haven't seen in your tests before.

PR #79, opened 2019-11-18
Make illegal values an explicitly tagged value.

ltratt:
Previously we used (sort of) null `GCBOX`'s as an illegal value. The problem with this is that we were haphazard in checking whether such values were ever used for anything *and* we only did so in debug mode. Any mistakes in their use were likely to end badly. This commit tentatively adds illegal values as an explicitly tagged entity: that means that we can't use them incorrectly, even in release mode. This probably slightly slows things down, but not enough that I can measure.


ptersilie:
Looks good.

bors r+

PR #80, opened 2019-11-18
Add Double divide operator for Ints, BigInts and Doubles.

ummarikar:
Add Double divide operator and include tests for all combinations. If `Int // Num` is an Int, then return an Int, otherwise return a Double.


ummarikar:
I can't get ```BigInt // Num``` to equal a ```Double``` if the answer is not a whole number. It always evaluates to ```BigInt```. Not sure if that is intended or if it is something to do with my code?

ltratt:
>I can't get BigInt // Num to equal a Double if the answer is not a whole number. It always evaluates to BigInt. Not sure if that is intended or if it is something to do with my code?

`f64` can only represent *some* numbers and the bigger the number the less likely it is to be able to represent it. For example, 50% of all numbers an `f64` can represent like between -1 and +1! So my guess is that you've chosen a number that `f64` can't represent accurately, and falls back to a nearby integer.

smarr:
@ummarikar did you check whether the chosen semantics conforms to any of the SOM's?
I didn't check, but if I remember correctly, my intention with a double div was that it **always** results in a `Double` to converge on a fixed type.

ummarikar:
Just checked - double div always results in a Double. Changes coming shortly!

ummarikar:
Double div now always returns a Double. Also included some divide by zero error tests.

ummarikar:
Only issue I had was coming up with appropriate test names. Please let me know of any better ones where applicable.

ltratt:
Please squash.

ummarikar:
Squashed.

PR #82, opened 2019-11-22
Add support for Integer class primitives.

ummarikar:
Add Integer as a Builtin and include a test to ensure `(Integer class) println.` prints `Integer`.

### DIFF
--- a/lang_tests/int_class.som
+++ b/lang_tests/int_class.som
@@ -1,0 +1,12 @@
+"
+VM:
+  status: success
+  stdout:
+    Integer
+"
+
+int_class = (
+    run = (
+        (Integer class) println.
+    )
+)
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -491,6 +491,7 @@ impl<'a> Compiler<'a> {
                         "false" => vm.instrs_push(Instr::Builtin(Builtin::False)),
                         "system" => vm.instrs_push(Instr::Builtin(Builtin::System)),
                         "true" => vm.instrs_push(Instr::Builtin(Builtin::True)),
+                        "Integer" => vm.instrs_push(Instr::Builtin(Builtin::Integer)),
                         _ => return Err(e),
                     },
                 }
--- a/src/lib/compiler/instrs.rs
+++ b/src/lib/compiler/instrs.rs
@@ -27,6 +27,7 @@ pub enum Instr {
 #[derive(Clone, Copy, Debug)]
 pub enum Builtin {
     False,
+    Integer,
     Nil,
     System,
     True,
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -113,6 +113,7 @@ pub struct VM {
     pub system_cls: Val,
     pub true_cls: Val,
     pub false_: Val,
+    pub int_: Val,
     pub nil: Val,
     pub system: Val,
     pub true_: Val,
@@ -155,6 +156,7 @@ impl VM {
             system_cls: Val::illegal(),
             true_cls: Val::illegal(),
             false_: Val::illegal(),
+            int_: Val::illegal(),
             nil: Val::illegal(),
             system: Val::illegal(),
             true_: Val::illegal(),
@@ -192,6 +194,7 @@ impl VM {
         vm.system_cls = vm.init_builtin_class("System", false);
         vm.true_cls = vm.init_builtin_class("True", false);
         vm.false_ = Inst::new(&vm, vm.false_cls.clone());
+        vm.int_ = Inst::new(&vm, vm.int_cls.clone());
         vm.system = Inst::new(&vm, vm.system_cls.clone());
         vm.true_ = Inst::new(&vm, vm.true_cls.clone());
 
@@ -299,6 +302,7 @@ impl VM {
                         Builtin::False => self.false_.clone(),
                         Builtin::System => self.system.clone(),
                         Builtin::True => self.true_.clone(),
+                        Builtin::Integer => self.int_.clone(),
                     });
                     pc += 1;
                 }
@@ -841,6 +845,7 @@ impl VM {
             system_cls: Val::illegal(),
             true_cls: Val::illegal(),
             false_: Val::illegal(),
+            int_: Val::illegal(),
             nil: Val::illegal(),
             system: Val::illegal(),
             true_: Val::illegal(),


ummarikar:
Hopefully, this is the right way to do it in yksom, if not, please shout! Otherwise, @ltratt was correct in saying this was not rocket science  (more like hide and seek). My initial assumption of the implementation was in fact exactly what I have submitted in this pull request here. But, instead of jumping straight in I wanted to be absolutely sure. After banging my head trying to understand Java SOM better, I found the following code in ```src/som/vm/Universe.java```:

```
setGlobal(symbolFor("nil"), nilObject);
setGlobal(symbolFor("true"), trueObject);
setGlobal(symbolFor("false"), falseObject);
setGlobal(symbolFor("system"), systemObject);
setGlobal(symbolFor("System"), systemClass);
setGlobal(symbolFor("Block"), blockClass);
```
I noticed in yksom the equivalent is in ```src/lib/compiler/instrs.rs```:
```
pub enum Builtin {
    False,
    Nil,
    System,
    True,
}
```
With this in mind I found the following in ```src/som/vm/Universe.java```:
```
initializeSystemClass(integerClass, objectClass, "Integer”);
…
public void initializeSystemClass(final SClass systemClass, final SClass superClass,
      final String name) {
    …
    // Insert the system class into the dictionary of globals
    setGlobal(systemClass.getName(), systemClass);
  }

```
The ```setGlobal``` call in the ```initializeSystemClass``` method, is exactly the call used to set the globals from before. And since the equivalent in yksom is the Builtin enum, I decided to look here to add Integer so we can implement the class primitive ```fromString```. I also included a test to ensure ```(Integer class) println.``` prints ```Integer```. All tests pass and everything seems to work fine. I thought I’d explain my thinking to justify my implementation.

ltratt:
As a temporary hack we can do this (or something like it). In the long run I *think* we might need a dictionary of globals (see `System::global` and so on). Perhaps that's even worth doing now -- have a look and see if you think you can do something here.

If we do stick with this PR's approach, then there's no point adding `int_` because that's simply a synonym for `int_cls` -- we only need the latter.

ltratt:
As a temporary hack we can do this (or something like it). In the long run I *think* we might need a dictionary of globals (see `System::global` and so on). Perhaps that's even worth doing now -- have a look and see if you think you can do something here.

If we do stick with this PR's approach, then there's no point adding `int_` because that's simply a synonym for `int_cls` -- we only need the latter.

ltratt:
Along these lines, yksom currently doesn't load classes entirely properly AFAICS: I think that's probably intimately connected with globals. Again that stuff is in the `System` class and might be worth looking at soon-ish.

ummarikar:
Sorry, I am not sure I understand. I found the SOM class for ```System``` but I can't find any Rust code implementing ```System::Global``` or, did you mean I have to create this? If so, could you point me to where I might put this?

smarr:
SOM's globals, which includes names bound to class objects like `Integer`, can be changed arbitrarily.

See: https://github.com/SOM-st/SOM/blob/master/Smalltalk/System.som#L29-L31

This means, there is typically a dictionary of globals. Some SOMs store a mutable cell in the dictionary to avoid a hash operation at the point where the global is accessed either lexically (e.g. `Integer println`) or via the operation (e.g. `(System global: #Integer) println`).
Generally it is assumed that `(System global: #Integer) == Integer` and so on holds.

ltratt:
@smarr Thanks for that, though these two statements might seem contradictory:

>SOM's globals, which includes names bound to class objects like Integer, can be changed arbitrarily.

vs.

>Generally it is assumed that (System global: #Integer) == Integer and so on holds.

I guess the question is partly whether changing the global called `Integer` should change anything about integer creation? I think your second statement implies "no" (which is what I would have guessed, but clarifying never hurts!).

smarr:
The following code is expected to work (minus typos):

```Smalltalk
(System global: #Integer) == Integer.   '==> true'
System global: #Integer put: 42.
(System global: #Integer) == Integer.   '==> true'
Integer println.                        '==> 42'
(System global: #Integer) println.      '==> 42'
```

Is this crazy? Yes it is. Is it unsafe? Yes. Absolutely.
Is this Smalltalk? Yes! Welcome to the world of uniform and simple languages, and the unexpected consequences.
Should a sane language allow this? If the answer is no, then the consequence is sacrificing uniformity. And, it's probably worth it.

And to answer your question about Integer creation: There are multiple possible interpretations, depending on the answer to whether you expect integer creation to rely on the global or not.
I would see those two as orthogonal concepts and integer creation does not use the `Integer` global, but instead knows the integer class independently of it. This breaks uniformity partially, but makes the system a little more robust.
Same applies to all object representations that have a literal syntax. Should they or should they not use the globals to find the classes. I'd answer no, at the cost of uniformity.

To me, globals are merely a convenient language-level namespace.

ltratt:
@smarr OK, that code sample is very helpful -- thanks!

@ummarikar I think we should close this PR and you should open a new PR that changes the `Global` bytecode so that it takes a string which is looked up dynamically at run-time (a bit like `Send`). This is quite different from what you've done up until now, but it won't require a *vast* amount of new code. For the first PR, don't worry about `System global:` if that seems like too much, as that can be done as a second PR. Sound like a plan?

ummarikar:
So in Java SOM there's a ```HashMap<SSymbol, SAbstractObject>``` in ```Universe.java``` which maps a symbol to the object that needs to be returned. And, when the Integer object is initialised its added to this HashMap. Is this what we need for Rust?

smarr:
Just as another example TruffleSOM uses an [Association](https://github.com/SOM-st/TruffleSOM/blob/master/src/trufflesom/vm/Universe.java#L85-L107) object: https://github.com/SOM-st/TruffleSOM/blob/master/src/trufflesom/vm/Universe.java#L786

ltratt:
> So in Java SOM there's a `HashMap<SSymbol, SAbstractObject>` in `Universe.java` which maps a symbol to the object that needs to be returned. And, when the Integer object is initialised its added to this HashMap. Is this what we need for yksom and we put this in `src/lib/vm/core.rs`? So essentially we have a HashMap of globals that we initialise and can change at runtime using `System global: #Integer put: 42.` for example.

We currently don't have a `Symbol` type, so we would probably need to add support for that first. Then we'd have `globals: HashMap<Symbol, Val>`, prepopulate it with the major globals (`Integer` etc.) and then implement the relevant `System` primitives for getting/setting. That's probably 3 PRs at least.

ummarikar:
Turns out I'm already implementing it in that way! I do have one question. In yksom would the ```Symbol``` type be an enum? Or would we make it its own standalone class?

ltratt:
`Symbol` would need to be its own struct (and, AFAICS, `Symbol` would more-or-less be a subset of `String_` in terms of its behaviour).

smarr:
`Symbol` is a **subclass** of `String`: https://github.com/SOM-st/SOM/blob/master/Smalltalk/Symbol.som#L26

ltratt:
@smarr Ah, now I see why @ummarikar made the `enum` comment.

@ummarikar What I would *probably* do to start with is add a field called something like `cls` to `String_` and change `String::new` to take a `bool` called something like `is_str` and then have:

```rust
if is_str {
    self.cls = vm.str_cls.clone();
} else {
    self.cls = vm.symbol_cls.clone();
}
```

Doing that will mean that wherever the VM expects a `String_` it will also work with a user-level `Symbol`, but any operations that `Symbol` overrides from `String` (i.e. the `SOM` class `String`, not the Rust struct `String_`) will work as expected.

ummarikar:
I know what to do now. Thank you, guys! Apologies for being slow

ltratt:
One other thing to add: we probably want to cache symbols, though we will have to differentiate them from strings (i.e. we'll need a separate cache for strings and symbols).

smarr:
Important for Symbols is that there's a global "cache"/table or similar mechanism usually. Symbols guarantee that pointer comparisons, i.e., `==` give `true` for Symbols with the same value (string).

I would not generally recommend that for Strings. String deduplication for arbitrary strings is a much more complex trade-off.

ltratt:
Closing in favour of #84.

PR #84, opened 2019-11-23
Add Symbol class.

ummarikar:
Add support for the som class - `Symbol`, a subclass of `String`. In the future, this should be used to implement classes as globals so we can support class primitives.

### DIFF
--- a/lang_tests/symbol.som
+++ b/lang_tests/symbol.som
@@ -1,0 +1,22 @@
+"
+VM:
+  status: success
+  stdout:
+    #ab
+    true
+    false
+    true
+    String
+    String
+"
+
+symbol = (
+    run = (
+        #ab println.
+        (#ab == #ab) println.
+        ((#a + #b) == #ab) println.
+        (('a' + 'b') asSymbol == #ab) println.
+        (('a' + #ab) class) println.
+        ((#ab + 'a') class) println.
+    )
+)
--- a/src/lib/vm/objects/string_.rs
+++ b/src/lib/vm/objects/string_.rs
@@ -13,6 +13,7 @@ use crate::vm::{
 #[derive(Debug, GcLayout)]
 pub struct String_ {
     s: String,
+    is_str: bool,
 }
 
 impl Obj for String_ {
@@ -21,7 +22,12 @@ impl Obj for String_ {
     }
 
     fn get_class(&self, vm: &VM) -> Val {
-        vm.str_cls.clone()
+        // FIXME This is a temporary hack until we sort out bootstrapping of the String_ class
+        if self.is_str {
+            vm.str_cls.clone()
+        } else {
+            vm.sym_cls.clone()
+        }
     }
 }
 
@@ -34,8 +40,8 @@ impl StaticObjType for String_ {
 }
 
 impl String_ {
-    pub fn new(vm: &VM, s: String) -> Val {
-        Val::from_obj(vm, String_ { s })
+    pub fn new(vm: &VM, s: String, is_str: bool) -> Val {
+        Val::from_obj(vm, String_ { s, is_str })
     }
 
     pub fn as_str(&self) -> &str {
@@ -57,6 +63,6 @@ impl String_ {
         let mut new = String::with_capacity(self.s.len() + other_str.s.len());
         new.push_str(&self.s);
         new.push_str(&other_str.s);
-        Ok(String_::new(vm, new))
+        Ok(String_::new(vm, new, true))
     }
 }
--- a/src/lib/vm/objects/string_.rs
+++ b/src/lib/vm/objects/string_.rs
@@ -29,6 +29,19 @@ impl Obj for String_ {
             vm.sym_cls.clone()
         }
     }
+
+    fn to_strval(&self, vm: &VM) -> Result<Val, Box<VMError>> {
+        Ok(String_::new(vm, self.s.to_string(), true))
+    }
+
+    fn ref_equals(&self, vm: &VM, other: Val) -> Result<Val, Box<VMError>> {
+        let other_str: &String_ = other.downcast(vm)?;
+
+        Ok(Val::from_bool(
+            vm,
+            (self.is_str == other_str.is_str) && (self.s == other_str.s),
+        ))
+    }
 }
 
 impl NotUnboxable for String_ {}
@@ -64,5 +77,9 @@ impl String_ {
         new.push_str(&self.s);
         new.push_str(&other_str.s);
         Ok(String_::new(vm, new, true))
+    }
+
+    pub fn to_symbol(&self, vm: &VM) -> Result<Val, Box<VMError>> {
+        Ok(String_::new(vm, self.s.to_string(), false))
     }
 }
--- a/src/lib/vm/val.rs
+++ b/src/lib/vm/val.rs
@@ -285,7 +285,11 @@ impl Val {
 
     pub fn to_strval(&self, vm: &VM) -> Result<Val, Box<VMError>> {
         match self.valkind() {
-            ValKind::INT => Ok(String_::new(vm, self.as_isize(vm).unwrap().to_string(), true)),
+            ValKind::INT => Ok(String_::new(
+                vm,
+                self.as_isize(vm).unwrap().to_string(),
+                true,
+            )),
             ValKind::GCBOX => self.tobj(vm).unwrap().to_strval(vm),
             ValKind::ILLEGAL => unreachable!(),
         }
--- a/src/lib/vm/val.rs
+++ b/src/lib/vm/val.rs
@@ -285,7 +285,7 @@ impl Val {
 
     pub fn to_strval(&self, vm: &VM) -> Result<Val, Box<VMError>> {
         match self.valkind() {
-            ValKind::INT => Ok(String_::new(vm, self.as_isize(vm).unwrap().to_string())),
+            ValKind::INT => Ok(String_::new(vm, self.as_isize(vm).unwrap().to_string(), true)),
             ValKind::GCBOX => self.tobj(vm).unwrap().to_strval(vm),
             ValKind::ILLEGAL => unreachable!(),
         }
@@ -652,7 +652,7 @@ mod tests {
         let vm = VM::new_no_bootstrap();
 
         let v = {
-            let v = String_::new(&vm, "s".to_owned());
+            let v = String_::new(&vm, "s".to_owned(), true);
             let v_tobj = v.tobj(&vm).unwrap();
             let v_int: &dyn Obj = v_tobj.deref().deref();
             let v_recovered = Val::recover(v_int);
@@ -667,7 +667,7 @@ mod tests {
     #[test]
     fn test_cast() {
         let vm = VM::new_no_bootstrap();
-        let v = String_::new(&vm, "s".to_owned());
+        let v = String_::new(&vm, "s".to_owned(), true);
         assert!(v.downcast::<String_>(&vm).is_ok());
         assert_eq!(
             *v.downcast::<Class>(&vm).unwrap_err(),
@@ -681,7 +681,7 @@ mod tests {
     #[test]
     fn test_downcast() {
         let vm = VM::new_no_bootstrap();
-        let v = String_::new(&vm, "s".to_owned());
+        let v = String_::new(&vm, "s".to_owned(), true);
         assert!(v.downcast::<String_>(&vm).is_ok());
         assert!(v.downcast::<Class>(&vm).is_err());
         assert!(v.try_downcast::<String_>(&vm).is_some());
--- a/lib/SOM/String.som
+++ b/lib/SOM/String.som
@@ -2,5 +2,6 @@ String = (
     concatenate: argument = primitive
     + argument = ( ^self concatenate: argument asString )
     asString = (^self)
+    asSymbol = primitive
     print = ( system printString: self )
 )
--- a/lib/SOM/Symbol.som
+++ b/lib/SOM/Symbol.som
@@ -1,4 +1,6 @@
 Symbol = String (
     asString = primitive
     asSymbol = ( ^self )
+    "FIXME implement super keyword instead"
+    print    = ( '#' print. (self asString) print )
 )
--- a/lib/SOM/Symbol.som
+++ b/lib/SOM/Symbol.som
@@ -1,0 +1,4 @@
+Symbol = String (
+    asString = primitive
+    asSymbol = ( ^self )
+)
--- a/src/lib/compiler/ast.rs
+++ b/src/lib/compiler/ast.rs
@@ -66,5 +66,6 @@ pub enum Expr {
     },
     Return(Box<Expr>),
     String(Lexeme<StorageT>),
+    Symbol(Lexeme<StorageT>),
     VarLookup(Lexeme<StorageT>),
 }
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -55,6 +55,7 @@ impl<'a> Compiler<'a> {
                     "Block" => Some(vm.block_cls.clone()),
                     "Boolean" => Some(vm.bool_cls.clone()),
                     "nil" => None,
+                    "String" => Some(vm.str_cls.clone()),
                     _ => unimplemented!(),
                 };
             } else {
@@ -109,7 +110,7 @@ impl<'a> Compiler<'a> {
         }
 
         Ok(Class {
-            name: String_::new(vm, name),
+            name: String_::new(vm, name, true),
             path: compiler.path.to_path_buf(),
             supercls,
             num_inst_vars: astcls.inst_vars.len(),
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -235,6 +235,7 @@ impl<'a> Compiler<'a> {
                 }
                 "sqrt" => Ok(MethodBody::Primitive(Primitive::Sqrt)),
                 "asString" => Ok(MethodBody::Primitive(Primitive::AsString)),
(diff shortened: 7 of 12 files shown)


ummarikar:
```thread 'main' panicked at 'internal error: entered unreachable code', src/lib/vm/val.rs:154:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.```
This is the error I keep getting, what am I missing?

**Comment by ummarikar** [comment]
This _fixes_ the previous error. But, it's up to you @ltratt to decide whether this was a fix or a hack! This solution came to me as a way to get rid of the error, but not a way to get it to work in the way you previously suggested it should work. But, I thought I'd commit this anyway just in case you might be happy with it.

**Comment by ltratt** [comment]
Did you check line 154 of `val.rs`?

**Comment by smarr** [comment]
In SOM, the following should be true (untested, may have typos):

```Smalltalk
#ab == #ab.                  '==> true'
(#a + #b) == #ab.            '==> true'
('a' + 'b') asSymbol == #ab. '==> true'
```

ummarikar:
On line 154: `ValKind::ILLEGAL => unreachable!(),` In the downcast method, I am not sure why.

ummarikar:
> In SOM, the following should be true (untested, may have typos):
> 
> ```smalltalk
> #ab == #ab.                  '==> true'
> (#a + #b) == #ab.            '==> true'
> ('a' + 'b') asSymbol == #ab. '==> true'
> ```

Not forgetting this don't worry! Just trying to get what I've got to run before I send more commits here. I should have been clearer about that sorry

ltratt:
> On line 154: ValKind::ILLEGAL => unreachable!(), In the downcast method, I am not sure why.

Ah, this almost always indicates a bootstrapping error. What's probably happening is that some objects are creating `String_`'s before `str_cls` is initialised, so they end up with a reference to `ValKind::ILLEGAL` instead. This can be fixed, though it is slightly annoying to do -- I'll walk you through how to do that in person. For the time being, your fix here is OK, but please add a comment in the `get_class` function along the lines of `XXX This is a temporary hack until we sort out bootstrapping of the String_ class` so that we remember to go back and fix this properly.

smarr:
> Not forgetting this don't worry!

@ummarikar  just wanted to assist with some test cases :)

ummarikar:
Would we need to add the ability to write symbols as ```#ab``` for example to som.l and som.y?

ltratt:
> Would we need to add the ability to write symbols as #ab for example to som.l and som.y?

Hint: have you checked those files? If so, I think you'll find the answer to your question!

ummarikar:
This is what I got so far until I ran into a blocker so I decided to commit it so @ltratt can decide whether it is correct and where I could potentially be going down the wrong path. I implemented what I assumed to be what you meant by caching `String` objects (for `Symbol` objects). Then various features like `asString` and `asSymbol` to (_try to_) get the test that @smarr kindly provided us to work. Some things may be missing because I was experimenting with my understanding rather than producing a perfect solution straight away. I am starting to struggle so hopefully the experts can shed some light!

lang_tests/symbol.som
```diff
@@ -0,0 +1,16 @@
+"
+VM:
+  status: success
+  stdout:
+    true
+    false
+    true
```
ummarikar: @smarr I ran this in Java SOM and got this instead of true, true, true. Is this the intended behaviour?

lang_tests/symbol.som
```diff
@@ -0,0 +1,16 @@
...
+"
+
+symbol = (
+    run = (
+        (#ab == #ab) println.
+        ((#a + #b) == #ab) println.
+        (('a' + 'b') asSymbol == #ab) println.
```
ummarikar: This line returns false even though when I tested `('a' + 'b') asSymbol println.` and `#ab println.` they both printed `#ab` and they both return a type `Symbol`. Not sure why this line is false then.

ltratt:
You'll need to tell us what the problem is if you want help on solving it!

lang_tests/symbol.som
smarr: Hm, this is unfortunate. I would expect that concatenation preserves the type. But I suppose, it is currently only implemented for `String` objects.

lang_tests/symbol.som
smarr: You will need some mechanism to make sure there is only ever one instance of a symbol, or that all identity comparisons return true.
Typically, one would have a hash table or so with all symbol objects keyed on the string.

lang_tests/symbol.som
ltratt: What is the expected semantics of the result of, for example, `Symbol + String`? Is it `Symbol`?

lang_tests/symbol.som
ltratt: Overriding identity comparisons in yksom is (deliberately!) easy (via the `ref_equals` method), so we don't want to cache arbitrary symbols. We do, IMHO, want to cache constant symbols though.

lang_tests/symbol.som
smarr: In Pharo, all concatenations result in a String. This is a little surprising, because there was quite a bit of effort spent on making sure that collections would maintain their type on operations that combine collections, or may result in new ones. From a design perspective, this seems a very similar example.

Not sure what the practical consequences are of ensuring the type of the receiver. I guess, for the moment, it is safer to go with SOM-java semantics.

ltratt: OK. @ummarikar we'll need to add tests that ensure that Symbol + String (and String + Symbol) result in a String, so that if this changes in the future we catch the right bits of code to change!

ummarikar:
Done thanks for the help @ltratt and @smarr !

src/lib/vm/objects/mod.rs
```diff
@@ -80,6 +80,11 @@ pub trait Obj: std::fmt::Debug + abgc::GcLayout {
         unimplemented!();
     }
 
+    /// Convert this object to a `Val` that represents a SOM symbol.
+    fn to_symbol(&self, _: &VM) -> Result<Val, Box<VMError>> {
```
ltratt: Since this operation is only valid on `String_`s, I think this should not be in `Obj` -- it should be solely in `String_`.

src/lib/vm/objects/string_.rs
```diff
@@ -22,6 +22,8 @@ use crate::vm::{
 #[derive(Debug, GcLayout)]
 pub struct String_ {
     s: String,
+    as_string: String,
```
ltratt: This field name seems a bit weird (it sounds like a function). I'd stick with `s` myself unless there's a good reason to change.

src/lib/vm/objects/string_.rs
```diff
@@ -43,12 +67,29 @@ impl StaticObjType for String_ {
 }
 
 impl String_ {
-    pub fn new(vm: &VM, s: String) -> Val {
-        Val::from_obj(vm, String_ { s })
+    pub fn new(vm: &VM, s: String, is_str: bool) -> Val {
+        let as_string;
```
ltratt: This is inefficient because we always copy the `String` coming in. Can the caller (lexer / parser?) not have added `#` where appropriate?

src/lib/vm/val.rs
```diff
@@ -287,12 +287,28 @@ impl Val {
...
+                vm,
+                self.as_isize(vm).unwrap().to_string(),
+                true,
+            )),
             ValKind::GCBOX => self.tobj(vm).unwrap().to_strval(vm),
             ValKind::ILLEGAL => unreachable!(),
         }
     }
 
+    pub fn to_symbol(&self, vm: &VM) -> Result<Val, Box<VMError>> {
```
ltratt: Ditto to the `Obj` comment: this function almost certainly shouldn't exist.

src/lib/vm/objects/string_.rs
ummarikar: This is a field I use for printing. So a `Symbol` which is `#ab` will have `as_string = '#ab'` and `s = 'ab'`. Whereas, a `String` which is `ab` will have `as_string = 'ab'` and `s = 'ab'`. This is so that we get that difference in behaviour for printing. If you see `as_str` in `String_`, `as_String` is returned. So when this method is called, a `Symbol` will return `'#ab'` and a `String` will return `'ab'` even if both of their `s`'s are the same.

ltratt: Ah, sorry, I thought you'd deleted `s`. My mistake.

Anyway, we can't add an extra field here: it's too costly (it doubles the cost of every string!). You'll need to find another way of doing this dynamically :)

src/lib/vm/objects/string_.rs
ummarikar: I did this so we wouldn't have to change the functionality of concatenating `Symbol` objects.

src/lib/vm/objects/string_.rs
smarr: SOM uses normally a `print` method: https://github.com/SOM-st/SOM/blob/master/Smalltalk/Symbol.som#L33
```Smalltalk
print    = ( '#' print. super print )
```

ummarikar: I tried this but it didn't like the super keyword. I'll make sure to add support for this then.

ummarikar:
I've included a temporary fix for printing `Symbol`. I think I should do `super` in a separate PR if everyone is okay with that.

ltratt:
The problem with temporary fixes is that they often end up being not-very-temporary! In this case, because `super` is probably not *too* difficult, I would suggest doing a separate PR for `super`, then coming back and finishing off this PR. They're pretty separate, so the code here shouldn't bit-rot.

ummarikar:
I'm struggling with `super` not sure where to start with this. I'm guessing I'd make it a `Builtin` but I don't know how I'd 1. get the object 2. convert it to its super type.

smarr:
@ummarikar `super` traditionally has support in the parser. Since super starts the lookup for a method at the super class of the class in which the method with the send operation is defined, it is not sufficient to simply take the super class at run time.

smarr:
Hm, no, I am mistaken. SOM-java actually does all the lookup at run time, because it as a reference to the method object, in which the `super` was, and the method refers to the correct class, from which it can obtain the super class.

ltratt:
Does SOM++ do the same as SOM-java?

smarr:
Basically yes. The code is slightly different, but that might be because the "method holder" (the class with the lexical definition of the method/block) is not set properly/consistently:

See
https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Interpreter.java#L139

and
https://github.com/SOM-st/SOMpp/blob/master/src/interpreter/Interpreter.cpp#L438-L441

smarr:
TruffleSOM works slightly differently.
Here the parse encodes the class in the AST, which is then used for the look up on first execution:
https://github.com/SOM-st/TruffleSOM/blob/master/src/trufflesom/interpreter/nodes/dispatch/SuperDispatchNode.java#L52-L59

ltratt:
I guess that since a class's superclass is determined at the point that it's created that it can't be changed and thus TruffleSOM's optimisation is entirely transparent to the user?

smarr:
SOM used to have represent everything, including superclass relationships as fields, which could be updated. This is currently not the case anymore.

Currently, superclasses are determined at class-load time.

ltratt:
> SOM used to have represent everything, including superclass relationships as fields, which could be updated. This is currently not the case anymore.

What changed? Was it to make it easier for TruffleSOM to optimise or ...? Just curious!

I'm also not entirely sure of the semantics of `super`: my working guess is that `super m` means "call method `m` on my superclass with `self` set to me" and that I can do things like `super super m` if I want to skip to my supersuperclass? I assume that means that `super` is a special keyword of sorts?

smarr:
> What changed? Was it to make it easier for TruffleSOM to optimise or ...? Just curious!

It was me learning, more than anything else. And I wanted to avoid complexity and strangeness that I didn't find necessary.
With my current knowledge, it was an unnecessary change.

The optimizations in TruffleSOM wouldn't be any more complicated. And there is not efficiency wi by the change.
All it takes is invalidating an [Assumption](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/Assumption.html), or writing to a quasi-immutable in RPython when the superclass changes.

The basic are clear, I guess, but we got a few extra bits here, too:
```Smalltalk
A = (
  hello = ( ^ #world )
)

B = A (
  hello = ( ^ #welt )
  superHello = ( ^ super hello )
  lookupIsBoundStatically = ( ^ super superHello )
)

C = B (
  supersuper = ( super super )
)

C new hello. "==> #welt"
C new superHello. "==> #world"
C new lookupIsBoundStatically "==> should result in a #doesNotUnderstand error saying superHello not found in A"
C new supersuper "==> should result in a #doesNotUnderstand error saying super not found in B"
```

And no, `super` can not be chained. `super` is merely an alias for `self`, but binds the start of the lookup.
Skipping super classes is not supported.

The error messages in SOM-java will not be as good as they could be.

ltratt:
> It was me learning, more than anything else. And I wanted to avoid complexity and strangeness that I didn't find necessary. With my current knowledge, it was an unnecessary change.

OK, so it sounds like it would be more SOM-esque to make this fully dynamic and assume that TruffleSOM (and any others?) will move to fully dynamic `super` lookups in the future?

> super is merely an alias for self, but binds the start of the lookup.

How is that communicated to the runtime? Is there a special bytecode or something?

@ummarikar In terms of this PR, since Stefan and I are still debating how `super` works, I think we *should* get this PR merged, even though it would be better with support for `super`. So, please squash (and remember that "squashing" doesn't have to mean "squash down to one commit" -- it means "if you want to take this opportunity to reduce the number of commits and/or rewrite commit messages, please do so").

smarr:
> TruffleSOM (and any others?) will move to fully dynamic super lookups in the future?

I am not sure what this is supposed to mean.
In the current version of SOM super classes cannot be changed. That's fact. Whether that is going to change, I don't know.
The phrase "fully dynamic" seems unhelpful to me given possible optimizations which will avoid unnecessary run-time lookups.

> How is that communicated to the runtime? Is there a special bytecode or something?
Yes: https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Bytecodes.java#L47

And no, from my point of view there's no debate here.
There is how SOM currently works, and then there is how SOM might work if someone changes it.
In either case, Smalltalk implementations that cache lookups always go and invalidate any kind of lookup caches on class hierarchy changes. Super is no exception and changing a super class is assumed to be extremely rare. So, the goal is to avoid any overhead on the normal execution path, but allow the case where something changes to be slow.

ltratt:
>>   TruffleSOM (and any others?) will move to fully dynamic super lookups in the future?
> I am not sure what this is supposed to mean.

At the moment different SOMs do quite different optimisations based in part on different interpretations of the semantics. I'm still trying to work out in my head which are acceptable and which aren't. I'm getting there, though, I think!

ummarikar:
@ltratt still happy for me to squash?

ltratt:
@ummarikar Yes.

ummarikar:
Squashed.

ltratt:
Did you rebase this against origin/master? If not, please do so, and force push.

ummarikar:
Didn't rebase before the merge attempt. Just rebased now, getting many errors I don't recognise having a look at them now.

ummarikar:
Screenshot 2019-11-27 at 17 27 07
Getting these errors not sure why because I haven't made any changes here

ltratt:
Try `cargo update` then `cargo check`.

ummarikar:
Screenshot 2019-11-27 at 17 39 31
Now this is happening. (Errors are gone though thank you!)

ltratt:
There's no way I'll be able to debug that one remotely! I assume when you rebased you had some merge conflicts? If so, my guess is that you resolved one of them incorrectly.

ummarikar:
> There's no way I'll be able to debug that one remotely! I assume when you rebased you had some merge conflicts? If so, my guess is that you resolved one of them incorrectly.

Okay, now that I know this is probably me, I'll have a look!

ummarikar:
Force pushed as I can't find the issue in the changes I made. Can you spot anything?

ltratt:
You get to debug this one all by yourself!

ummarikar:
I'm having the same issue in `softdevteam:master` I don't think it's any of my changes. But also wondering if it's something to do with my computer.

ltratt:
It probably is your machine (it works fine for me on multiple OpenBSD / Linux mcahines). Try updating your nightly etc.

ummarikar:
Updated, uninstalled, reinstalled. Tried everything nothing worked not even commits that have worked on my machine before work now.

ummarikar:
FYI I'm using a mac

ltratt:
I'm afraid I can't debug your machine. You either need to work out a) what's gone wrong on your machine b) what's wrong about yksom (or some dependency) that's suddenly triggered something on your machine. There's little, if anything, I can do to help I'm afraid.

ummarikar:
Don't worry I understand. Just wanted to keep you updated on the situation.

ummarikar:
@ltratt thanks for your help you were right about `lang_tester` version `0.3.3` works on mac and it stops working from `0.3.4` onwards. I ran the tests with version `0.3.3` and they pass.

ummarikar:
bors try

ltratt:
OK, can you bisect the precise commit from lang_tester that causes a problem and file an issue against that repository? `git bisect` is your friend here (as is `lang_tester = { path = "/path/to/lang_tester" }` in your `Cargo.toml`). Thanks!

ummarikar:
> OK, can you bisect the precise commit from lang_tester that causes a problem and file an issue against that repository? `git bisect` is your friend here (as is `lang_tester = { path = "/path/to/lang_tester" }` in your `Cargo.toml`). Thanks!

Will do! And thanks again this is such a relief!

ltratt:
Please squash.

ummarikar:
Squashed.

PR #86, opened 2019-11-25
Move the UnsafeCell from SOMStack to VM.

ltratt:
This is how I should have done things in the first place, but I'd forgotten one
of my guiding principles, which is that `VM` should (roughly) model what a
multi-threaded interpreter should look like. In that model, some sort of lock
would be need to held before `SOMStack` can be accessed, but after that
`SOMStack` should be fairly normal.

This commit thus makes VM store `SOMStack` in an `UnsafeCell`, which means we
can then remove the inner `UnsafeCell` from `SOMStack` itself.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -86,6 +86,10 @@ macro_rules! stry {
     }};
 }
 
+/// The core VM struct. Although SOM is single-threaded, we roughly model what a multi-threaded VM
+/// would need to look like. That is, since this struct would need to be shared between threads and
+/// called without a single lock, thread-safety would need to be handled internally. We model that
+/// with [`UnsafeCell`].
 pub struct VM {
     classpath: Vec<String>,
     pub block_cls: Val,
@@ -112,7 +116,7 @@ pub struct VM {
     /// reverse_sends is an optimisation allowing us to reuse sends: it maps a send `(String,
     /// usize)` to a `usize` where the latter represents the index of the send in `sends`.
     reverse_sends: UnsafeCell<HashMap<(String, usize), usize>>,
-    stack: SOMStack,
+    stack: UnsafeCell<SOMStack>,
     strings: UnsafeCell<Vec<Val>>,
     /// reverse_strings is an optimisation allowing us to reuse strings: it maps a `String to a
     /// `usize` where the latter represents the index of the string in `strings`.
@@ -152,7 +156,7 @@ impl VM {
             instrs: UnsafeCell::new(Vec::new()),
             sends: UnsafeCell::new(Vec::new()),
             reverse_sends: UnsafeCell::new(HashMap::new()),
-            stack: SOMStack::new(),
+            stack: UnsafeCell::new(SOMStack::new()),
             strings: UnsafeCell::new(Vec::new()),
             reverse_strings: UnsafeCell::new(HashMap::new()),
             frames: UnsafeCell::new(Vec::new()),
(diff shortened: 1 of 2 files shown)


smarr:
> some sort of lock would be need to held before `SOMStack` can be accessed

This is strange. What exactly does `SOMStack` contain?
It says it is a "A fixed size stack of SOM values".

In all SOM implementations I am aware of, this would correspond to a single method/block activation.
While I can imaging that multiple threads share a method/block activations, I would hope that there's no need for a lock. Normally, multiple threads should only access the part that is accessible using lexical addressing (fixed stack offsets).

Given that all SOM implementations use a separate stack for a single method activation, seeing the stack in the VM is also unexpected. SOM implementations tent to have a `frame` field in the VM or interpreter. Frames then link to the previous frame.

Having a single stack seems to invite a lot of complexity with blocks/closures (i.e., manage stack regions that remain reachable/alive/in use after a method returned.

ltratt:
A `SOMStack` is a stack of `Val`s i.e. roughly "a stack of pointers to objects". AFAIK SOM is single-threaded, so `yksom` doesn't do anything clever with locks and so on (`UnsafeCell` is a weird concept if you're not used to Rust, so I wouldn't worry too much about its semantics). Also AFAIK SOM doesn't have any support for continuations/co-routines (or similar), so there's no problem having a single global stack. Please tell me if I'm wrong!

smarr:
The multithread stuff is just a distraction here. Let's ignore it.

```Smalltalk
ClosuresAreValuesAndObjectsAndScopes = (
  | get inc |
  

  makeCounter = (
    | cnt result reserve1 reserve2 reserve3 |
    cnt := 0.
    
    get := [ cnt ].
    inc := [ cnt := cnt + 1 ].
  )
  
  justUseSomeSpace = (
    | reserve1 reserve2 reserve3 reserve4 reserve5 |
    "I don't do anything but consume and relase stack space"
    reserve1 println. " ==> nil "
    reserve2 println. " ==> nil "
    reserve3 println. " ==> nil "
    reserve4 println. " ==> nil "
    reserve5 println. " ==> nil "
  )

  run = (
    self makeCounter.
    self justUseSomeSpace.
    
    get value println. " ==> 0 "

    inc value.
    self justUseSomeSpace.
    get value println. " ==> 1 "

    self justUseSomeSpace.

    inc value.
    get value println. " ==> 2 "
  )
)
```

should print:

```
SOM-java$ ./som.sh -cp core-lib/Smalltalk ClosuresAreValuesAndObjectsAndScopes.som            ✹ ✭master
nil
nil
nil
nil
nil
0
nil
nil
nil
nil
nil
1
nil
nil
nil
nil
nil
2
```

So, where in memory is `cnt`? I'd expect it to be on a "stack". In standard SOM terminology, it would be in a `frame`.

ltratt:
yksom prints out exactly the same. I think I now see the confusion. At the moment yksom doesn't store variable on the stack: they're heap allocated, so that escaping closures like your example work correctly. The stack is purely used for push/pop style operations, which are separate from variables.

Of course, this is very inefficient, since *most* variables do have stack-compatible lifetimes.

smarr:
I see.
Seems like such a design would ensure maximum fun for debugging.

ltratt:
Which design?

smarr:
The design of separating variables, arguments, and stack space.
Smalltalks classically keep them close together (called a Context, or stack frame) so that it is easy to make sense of what the interpreter is doing, especially when debugging bytecode or primitive implementations. (It's not unheard of to miss a push or pop here and there. Great fun to debug such issues.)
At least conceptually, most other languages seem to chose a very similar, conservative design.

If I understand correctly, in your design, you don't even know which method uses a specific stack slot, much less what content it should have. Sounds like "fun" to me. But perhaps I am too pessimistic.

ltratt:
Yes, you are too pessimistic :) This is an extremely easy way of implementing things. Whether it's any easier than other mechanisms or not, I'm not sure, but it's definitely one of the easier parts of the implementation.

smarr:
I am not talking about implementing things. I am talking about making sense of a system that is execution >10k bytecodes to reach a state in which it crashes. And then debugging the crash :)

ltratt:
I've never had a problem with this way of laying things out TBH.

ptersilie:
Since the discussion has settled down now:

bors r+

PR #90, opened 2020-01-15
Add support for globals.

ummarikar:
Add a hashmap which maps symbols to their corresponding object and retrieve an object from this hashmap when its symbol is referenced during runtime.

### DIFF
--- a/lang_tests/system_global.som
+++ b/lang_tests/system_global.som
@@ -1,0 +1,36 @@
+"
+VM:
+  status: success
+  stdout:
+    true
+    true
+    42
+    42
+    String
+    21
+    21
+    nil
+    true
+    false
+"
+
+system_global = (
+    run = (
+        ((system global: #Integer) == Integer) println.
+        system global: #Integer put: 42.
+        ((system global: #Integer) == Integer) println.
+        Integer println.
+        (system global: #Integer) println.
+        system global: #Integer put: 'a'.
+        Integer class println.
+        system global: #ab put: 21.
+        (system global: #ab) println.
+        ab println.
+        system global: #nil put: 'a'.
+        system global: #true put: 'b'.
+        system global: #false put: 'c'.
+        nil println.
+        true println.
+        false println.
+    )
+)
--- a/lang_tests/system_global_invalid_symbol_err.som
+++ b/lang_tests/system_global_invalid_symbol_err.som
@@ -1,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: InvalidSymbol
+"
+
+system_global_invalid_symbol_err = (
+    run = (
+        system global: #ab.
+    )
+)
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -238,6 +238,8 @@ impl<'a> Compiler<'a> {
                 "asSymbol" => Ok(MethodBody::Primitive(Primitive::AsSymbol)),
                 "class" => Ok(MethodBody::Primitive(Primitive::Class)),
                 "concatenate:" => Ok(MethodBody::Primitive(Primitive::Concatenate)),
+                "global:" => Ok(MethodBody::Primitive(Primitive::Global)),
+                "global:put:" => Ok(MethodBody::Primitive(Primitive::GlobalPut)),
                 "halt" => Ok(MethodBody::Primitive(Primitive::Halt)),
                 "hashcode" => Ok(MethodBody::Primitive(Primitive::Hashcode)),
                 "inspect" => Ok(MethodBody::Primitive(Primitive::Inspect)),
@@ -485,13 +487,20 @@ impl<'a> Compiler<'a> {
                             vm.instrs_push(Instr::VarLookup(depth, var_num));
                         }
                     }
-                    Err(e) => match self.lexer.lexeme_str(&lexeme) {
-                        "nil" => vm.instrs_push(Instr::Builtin(Builtin::Nil)),
-                        "false" => vm.instrs_push(Instr::Builtin(Builtin::False)),
-                        "system" => vm.instrs_push(Instr::Builtin(Builtin::System)),
-                        "true" => vm.instrs_push(Instr::Builtin(Builtin::True)),
-                        _ => return Err(e),
-                    },
+                    Err(_) => {
+                        let lex_string = self.lexer.lexeme_str(&lexeme);
+
+                        match lex_string {
+                            "nil" => vm.instrs_push(Instr::Builtin(Builtin::Nil)),
+                            "false" => vm.instrs_push(Instr::Builtin(Builtin::False)),
+                            "true" => vm.instrs_push(Instr::Builtin(Builtin::True)),
+                            _ => {
+                                vm.instrs_push(Instr::Global(
+                                    vm.add_symbol(lex_string.to_string()),
+                                ));
+                            }
+                        }
+                    }
                 }
                 Ok(1)
             }
--- a/src/lib/compiler/instrs.rs
+++ b/src/lib/compiler/instrs.rs
@@ -2,6 +2,7 @@
 pub enum Instr {
     Block(usize),
     Builtin(Builtin),
+    Global(usize),
     ClosureReturn(usize),
     Double(f64),
     InstVarLookup(usize),
@@ -36,6 +37,8 @@ pub enum Primitive {
     Div,
     DoubleDiv,
     Equals,
+    Global,
+    GlobalPut,
     GreaterThan,
     GreaterThanEquals,
     Halt,
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -44,6 +44,8 @@ pub enum VMError {
         expected: ObjType,
         got: ObjType,
     },
+    /// Tried to access a global before it being initialised.
+    InvalidSymbol,
     /// Tried to do a shl or shr with a value below zero.
     NegativeShift,
     /// A specialised version of TypeError, because SOM has more than one number type (and casts
@@ -111,6 +113,7 @@ pub struct VM {
     pub system: Val,
     pub true_: Val,
     blockinfos: UnsafeCell<Vec<BlockInfo>>,
+    globals: UnsafeCell<HashMap<usize, Val>>,
     inline_caches: UnsafeCell<Vec<Option<(Val, Gc<Method>)>>>,
     instrs: UnsafeCell<Vec<Instr>>,
     sends: UnsafeCell<Vec<(String, usize)>>,
@@ -156,6 +159,7 @@ impl VM {
             system: Val::illegal(),
             true_: Val::illegal(),
             blockinfos: UnsafeCell::new(Vec::new()),
+            globals: UnsafeCell::new(HashMap::new()),
             inline_caches: UnsafeCell::new(Vec::new()),
             instrs: UnsafeCell::new(Vec::new()),
             sends: UnsafeCell::new(Vec::new()),
@@ -167,7 +171,6 @@ impl VM {
             reverse_symbols: UnsafeCell::new(HashMap::new()),
             frames: UnsafeCell::new(Vec::new()),
         };
-
         // The very delicate phase.
         //
         // Nothing in this phase must store references to the nil object or any classes earlier
@@ -191,6 +194,10 @@ impl VM {
         vm.sym_cls = vm.init_builtin_class("Symbol", false);
         vm.system_cls = vm.init_builtin_class("System", false);
         vm.true_cls = vm.init_builtin_class("True", false);
+        unsafe { &mut *vm.globals.get() }.insert(
+            vm.add_symbol("system".to_string()),
+            Inst::new(&vm, vm.system_cls.clone()),
+        );
         vm.false_ = Inst::new(&vm, vm.false_cls.clone());
         vm.system = Inst::new(&vm, vm.system_cls.clone());
         vm.true_ = Inst::new(&vm, vm.true_cls.clone());
@@ -226,7 +233,14 @@ impl VM {
         let path = self
             .find_class(name)
             .unwrap_or_else(|_| panic!("Can't find builtin class '{}'", name));
-        self.compile(&path, inst_vars_allowed)
+
+        let val = self.compile(&path, inst_vars_allowed);
+        let idx = self.add_symbol(name.to_string());
+        unsafe { &mut *self.globals.get() }
+            .entry(idx)
+            .or_insert(val.clone());
+
+        val
     }
 
     /// Inform the user of the error string `error` and then exit.
@@ -324,6 +338,15 @@ impl VM {
                 }
                 Instr::Double(i) => {
                     unsafe { &mut *self.stack.get() }.push(Double::new(self, i));
+                    pc += 1;
+                }
+                Instr::Global(symbol_off) => {
+                    debug_assert!(unsafe { &*self.symbols.get() }.len() > symbol_off);
+                    if let Some(global) = unsafe { &mut *self.globals.get() }.get(&symbol_off) {
+                        unsafe { &mut *self.stack.get() }.push(global.clone());
+                    } else {
+                        return SendReturn::Err(Box::new(VMError::InvalidSymbol));
+                    }
                     pc += 1;
                 }
                 Instr::InstVarLookup(n) => {
@@ -476,6 +499,31 @@ impl VM {
                 unsafe { &mut *self.stack.get() }.push(stry!(
                     rcv.equals(self, unsafe { &mut *self.stack.get() }.pop())
                 ));
+                SendReturn::Val
+            }
+            Primitive::Global => {
+                let name = unsafe { &mut *self.stack.get() }.pop();
+                let as_string: &String_ = stry!(name.downcast(self));
+                let s = as_string.as_str();
+
+                if let Some(i) = unsafe { &mut *self.reverse_symbols.get() }.get(s) {
+                    if let Some(val) = unsafe { &mut *self.globals.get() }.get(&i) {
+                        unsafe { &mut *self.stack.get() }.push(val.clone());
+                        return SendReturn::Val;
+                    }
+                }
+                SendReturn::Err(Box::new(VMError::InvalidSymbol))
+            }
+            Primitive::GlobalPut => {
+                let value = unsafe { &mut *self.stack.get() }.pop();
+                let name = unsafe { &mut *self.stack.get() }.pop();
+                let as_string: &String_ = stry!(name.downcast(self));
+                let s = as_string.as_str();
+
+                if let Some(i) = unsafe { &mut *self.reverse_symbols.get() }.get(s) {
+                    unsafe { &mut *self.globals.get() }.insert(*i, value.clone());
+                    unsafe { &mut *self.stack.get() }.push(value);
+                }
                 SendReturn::Val
             }
             Primitive::GreaterThan => {
@@ -853,6 +901,7 @@ impl VM {
             system: Val::illegal(),
             true_: Val::illegal(),
             blockinfos: UnsafeCell::new(Vec::new()),
+            globals: UnsafeCell::new(HashMap::new()),
             inline_caches: UnsafeCell::new(Vec::new()),
             instrs: UnsafeCell::new(Vec::new()),
             sends: UnsafeCell::new(Vec::new()),
--- a/lang_tests/instance_vars2.som
+++ b/lang_tests/instance_vars2.som
@@ -2,11 +2,11 @@
 VM:
   status: error
   stderr:
-    ...line 11, column 11:
-      ...
-    Unknown variable 'x'
+   InvalidSymbol 
 "
 
 instance_vars2 = (
-    x = (^x)
+  run = (
+    x = 1
+  )
 )
--- a/lib/SOM/System.som
+++ b/lib/SOM/System.som
@@ -1,4 +1,6 @@
 System = (
-    printString: string = primitive
-    printNewline        = primitive
+    global: name            = primitive
+    global: name put: value = primitive
+    printString: string     = primitive
+    printNewline            = primitive
 )


ummarikar:
This PR is unfinished and is only "fully" implemented for the `Integer` class, just wanted to submit it so we can discuss my implementation before I finish the rest and subsequently squash. FYI the test passes.

src/lib/vm/core.rs
```diff
@@ -226,7 +228,17 @@ impl VM {
         let path = self
             .find_class(name)
             .unwrap_or_else(|_| panic!("Can't find builtin class '{}'", name));
-        self.compile(&path, inst_vars_allowed)
+
+        let val = self.compile(&path, inst_vars_allowed);
+        let idx = self.add_symbol(name.to_string());
+        let globals = unsafe { &mut *self.globals.get() };
+
+        if let Some(_) = globals.get(&idx) {
```
ltratt: `if globals.get(...).is_some()`

ltratt:
`yksom` currently doesn't handle globals in an ideal way in the sense that they can't be reassigned to (which, I think, SOM allows). Before we go down the full class primitives route, we should probably get the semantics of this right because we may then be able to merge / remove the `Builtin` opcode which currently does something very similar to what you've called `Globals`.

ummarikar:
How do you think we should do it?

ltratt:
I think you should make some test files that show what happens in other SOMs when you assign to a global, and then see how such behaviour might best be implemented in yksom.

ummarikar:
By assigning globals do you mean what was discussed here? https://github.com/softdevteam/yksom/pull/82#issuecomment-557651237

ltratt:
Yes, but there might be other ways of doing the same thing too (e.g. assigning to the variable `Integer` or `true` might work for all I know).

ummarikar:
Great I'll start that now and will let you know how I get on.

smarr:
@ummarikar please try all the things you can come up with, especially the things that should not work... I am, well, not entirely sure which cases are handled correctly...

Via `system` one can definitely change the globals. Using a simple assignment, I would hope not, but oh well...

ummarikar:
> @ummarikar please try all the things you can come up with, especially the things that should not work... I am, well, not entirely sure which cases are handled correctly...
> 
> Via `system` one can definitely change the globals. Using a simple assignment, I would hope not, but oh well...

Will do! Thanks

ummarikar:
Working on `global:put:` as I am typing this, just wanted to submit what I have done thus far so it can be critiqued.

ltratt:
I think we really need to see "hard" examples (as @smarr suggested) before critiquing the code, as the precise SOM semantics could change what we perceive to be the best route.

ummarikar:
Yes apologies, I’ve been working on something substantial but have hit a couple of roadblocks along the way. Will submit something either today or tomorrow.

ummarikar:
Still struggling to get this right don’t want to ask for help yet though so I will continue to try

lang_tests/system3.som
```diff
@@ -3,10 +3,13 @@ VM:
   status: success
   stdout:
     true
+    42
 "
 
 system2 = (
     run = (
         ((System global: #Integer) == Integer) println.
+        System global: #Integer put: 42.
```
ummarikar: Getting this error: thread 'main' panicked at 'assertion failed: !self.is_empty()', src/lib/vm/somstack.rs:55:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

src/lib/vm/core.rs
```diff
@@ -513,6 +513,20 @@ impl VM {
                 }
                 SendReturn::Val
             }
+            Primitive::GlobalPut => {
+                let stack = unsafe { &mut *self.stack.get() };
+                let value = stack.pop();
+                let name = stack.pop();
```
ummarikar: I think the issue mentioned in https://github.com/softdevteam/yksom/pull/90/commits/2236c10a69c6d82ca18c1c19ab941a0df489d049#r368585356 is because of the second pop. For some reason, the stack is empty when I am trying to get the name. This shouldn't be the case... should it?

ltratt: This is your code so... I dunno! ;)

ummarikar: `value` and `name` should both be on the stack but I can only get `value` I am not sure where `name` has gone every time I try to do a second pop it returns that stack is empty error. Still trying to figure out what the issue is, but I haven't made any progress yet.

src/lib/vm/core.rs
```diff
@@ -478,6 +499,34 @@ impl VM {
...
+                let reverse_symbols = unsafe { &mut *self.reverse_symbols.get() };
+                let globals = unsafe { &mut *self.globals.get() };
+                let as_string: &String_ = stry!(name.downcast(self));
+                let s = as_string.as_str();
+
+                if let Some(i) = reverse_symbols.get(s) {
+                    globals.insert(*i, value);
+                }
+                SendReturn::Val
```
ummarikar: Finally found the problem, not sure how I missed this. This line here means another`pop()` is being applied to the stack but there's nothing on it since we used `pop()` twice already.

ummarikar:
@smarr what are the intended semantics for `(System global: #Integer put: 42) println.`? Does global put return anything such as the value you give it (42 in this case)?

ummarikar:
Ignore me I had a look at the code for `global:put:` in som-java and referred to the comment you left here - https://github.com/softdevteam/yksom/pull/82#issuecomment-557651237.

src/lib/vm/core.rs
ummarikar: @ltratt since `global:put:` does not return anything what should I return here instead of `SendReturn::Val`? Is it `SendReturn::ClosureReturn`? If so, there is no implementation for this, for reference: https://github.com/softdevteam/yksom/blob/34c71cdc81ddba6693089264bc7ea5ce60018c8d/src/lib/vm/core.rs#L263
What is the purpose of `SendReturn::ClosureReturn` here? Would you like me to implement this?

smarr:
@ummarikar hm, I think this is a good question.
I believe it is common practice to return either `self` or the value from such mutating methods.

See for instance: `(Array new: 6) at: 4 put: 66`, which returns 66.

I would assume that returning the value is the most useful thing to do here,
given the `#at:put:` precedence, and that the value may be derived from an expression.
Returning `system` or `nil` seems not ideal.

ummarikar:
@smarr I agree! I’ll have ‘global:put:’ return ‘value’.

src/lib/vm/core.rs
```diff
@@ -478,6 +499,40 @@ impl VM {
...
+                let s = as_string.as_str();
+
+                if let Some(i) = reverse_symbols.get(s) {
+                    if let Some(v) = globals.get(i) {
+                        if v.get_class(self) == value.get_class(self) {
+                            globals.insert(*i, value.clone());
+                            stack.push(value.clone());
+                        }
+                    }
+                }
```
ummarikar: @ltratt I want to add tests for behaviour that shouldn't work for `global:put:`. Would we need to add to `VMError` for the following: if the given symbol cannot be found, if the given symbol is not a global and if the value provided is of the incorrect type?

ltratt: `TypeError` already handles the latter case, but you'll probably to add something new to `VMError` for the former case.

src/lib/vm/core.rs
```diff
@@ -478,6 +501,46 @@ impl VM {
...
+                    if let Some(v) = globals.get(i) {
+                        if v.get_class(self) == value.get_class(self) {
+                            globals.insert(*i, value.clone());
+                            stack.push(value.clone());
+                            return SendReturn::Val;
+                        } else {
+                            return SendReturn::Err(Box::new(VMError::TypeError {
+                                expected: v.dyn_objtype(self),
+                                got: value.dyn_objtype(self),
+                            }));
```
ummarikar: Trying `System global: #Integer put: 'a'.` returns `TypeError { expected: Inst, got: String_ }` still trying to find a way to get the object type from an `Instance` object.

ummarikar: Is there a way to get `ObjType::Int` from the `Integer` instance object? I think not because I could not find anything in the code so I am writing a function to do that but wanted to ask anyway.

ummarikar:
This is my `global` and `global:put:` solution. I have changed the name of the tests too so they are more descriptive.

ummarikar:
Just added the classes of som to `ast_to_instrs`.

ltratt:
Before I look into this in detail, can I check that you've run all the new tests on another SOM (e.g. Java SOM) to check that they behave exactly the same?

lang_tests/system_global.som
```diff
@@ -0,0 +1,19 @@
...
+    true
+    true
+    42
+    42
+"
+
+system_global = (
+    run = (
+        ((System global: #Integer) == Integer) println.
+        System global: #Integer put: 42.
```
smarr: Ok, so this is strange.
I would expect the test to be on `system` not `System`, which is the class.

I would expect what you got here to error, like this:

```
---> System global: #Integer

ERROR: Method global: not found in class System class
```

`system` is the instance, while `System` is the class.

```
---> system class == System
it = true
```

And:
```
---> system == (system global: #system)
it = true
```

lang_tests/system_global_invalid_symbol_err.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: InvalidSymbol
+"
+
+system_global_invalid_symbol_err = (
+    run = (
+        System global: #ab.
```
smarr: Hm, I think here I would not expect an error. And the error is also strangely ambiguous. I'd rather expect that `(system global: #ab) == nil`.

lang_tests/system_global_put_invalid_symbol_err.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: InvalidSymbol
+"
+
+system_global_put_invalid_symbol_err = (
+    run = (
+        System global: #ab put: 'a'.
```
smarr: Why does this error?

lang_tests/system_global_put_type_err.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: error
+  stderr: TypeError { expected: Int, got: String_ }
+"
+
+system_global_put_type_err = (
+    run = (
+        System global: #Integer put: 'a'.
```
smarr: Why does this error?
A type error here doesn't make sense. And I'd also not expect that an integer would work.

And, a thing for you and @ltratt to pounder: your error message talks about the concept `Type` which is not a thing in SOM. You got `Class` though, which conflates the notion of types and object factories.
And, well, there's no notion of `Int` and `String_` either. [of course this is nitpicking, sorry]

ummarikar:
I tried this in SOM but I got an error that `global:` and `global:put:` is not in the `System` class. Will try it in another SOM soon.

lang_tests/system_global_put_invalid_symbol_err.som
ummarikar: My understanding was that the globals were fixed, can a user enter any globals that they want? I may have misunderstood.

lang_tests/system_global.som
ummarikar: I got `System` from your comment here - https://github.com/softdevteam/yksom/pull/82#issuecomment-557651237 I assume you meant `system`?

lang_tests/system_global_put_invalid_symbol_err.som
ummarikar: By fixed I mean you have globals for your classes like `Integer` and `String` for example but you wouldn't have `ab` like this. Happy to change it though.

lang_tests/system_global.som
smarr: Sorry, my bad. I fixed the comment.

lang_tests/system_global_put_invalid_symbol_err.som
smarr:
```
---> system global: #ab
it = nil
---> ab

ERROR: Unable to resolve ab
$ ./som.sh -cp Smalltalk
SOM Shell. Type "quit" to exit.
---> system global: #ab put: 4
it = instance of System
---> ab
it = 4

Comment (smarr) [reply lang_tests/system_global_put_invalid_symbol_err.som]: And of course:

```
SOM Shell. Type "quit" to exit.

---> system global: #ab put: nil
it = instance of System
---> ab
it = nil
```

File: lang_tests/system_global_put_type_err.som (hunk 10 above)
Comment (ltratt) [reply lang_tests/system_global_put_type_err.som]: From memory, in Java SOM, if you try to e.g. use an integer where a string is required, you get `CastException`s in the Java code. So all I tried to do (as a temporary hack, at least) was make those language level errors. I'm sure there's a more SOM-esque way of doing this, but the current implementations didn't really seem to have thought of it, so I had to improvise...

Comment (ummarikar) [reply lang_tests/system_global_put_type_err.som]: So a user runs `system global: #Integer put: 'ab'.` then `Integer class.` will be a `String` ?

Comment (smarr) [reply lang_tests/system_global_put_type_err.som]:
```
---> system global: #Integer put: 'ab'
it = instance of System
---> Integer class
it = String
---> Integer == 4 class
it = false
```

**Comment by ummarikar** [comment]
Okay now that @smarr kindly, pointed out that I have been mistaking `System` for `system` I have run tests on java SOM and I believe I understand the semantics now. Please ignore my code above and I will start working on the changes.

File: src/lib/compiler/ast_to_instrs.rs [hunk 11]
```diff
@@ -485,13 +487,25 @@ impl<'a> Compiler<'a> {
...
+                    Err(e) => {
+                        let lex_string = self.lexer.lexeme_str(&lexeme);
+
+                        match lex_string {
+                            "nil" => vm.instrs_push(Instr::Builtin(Builtin::Nil)),
+                            "false" => vm.instrs_push(Instr::Builtin(Builtin::False)),
+                            "system" => vm.instrs_push(Instr::Builtin(Builtin::System)),
+                            "true" => vm.instrs_push(Instr::Builtin(Builtin::True)),
+                            _ => {
+                                if vm.is_global(lex_string) {
```
ummarikar: Instead of having a match here to match every possible global I've included a check to see if the given string is a global then if yes then it is pushed into the instructions stack. That way we can add new globals during runtime.

ltratt: Can other SOMs override things like `nil`, `system` and so on? If so, the way `Builtin` works is wrong, and you'll need to think about how to change it.

smarr:
```
SOM-java
system global: #system put: 54
--> freezes 
```

`nil` is fun:

```
---> system global: #nil put: 4
it = instance of System
---> nil
it = 4
---> nil ifNotNil: [ 'WAT?' println ]
WAT?
it = WAT?
```

This follows from "there are no keywords*", but simply well known globals...
Similarly:

```
---> system global: #true put: 45
it = instance of System
---> true
it = 45
```

I am not sure this is really how it should be.
I would be reasonably unconcerned to declare a few key literal values as unchangeable.

ltratt: If you tell us that's the semantics, then those are the semantics: but I do think we should not cheat relative to other SOMs (unless the intention is to change them in the near-ish future).

smarr: Well...

```
This is CSOM.
SOM Shell. Type "system exit" to exit.
---> system global: #system put: 54
it = instance of System
---> system
it = instance of System
---> system global: #nil put: 4
it = instance of System
---> nil
it = nil
---> system global: #true put: 45
it = instance of System
---> true
it = true
---> system global: #true
it = true
---> system global: #nil
it = nil
---> system global: #dd put: 45
it = instance of System
---> system global: #dd
it = 45
```

And:

```
This is SOM++
---> system global: #system put: 54
Segmentation fault

This is SOM++
---> system global: #nil put: 4
it = instance of System
---> nil
it = 4
---> system global: #true put: 45
it = instance of System
---> true
it = 45
```

So, do we go here by majority?
I think it makes sense to let `true`, `false`, `nil` be special. For `system` I'd think it should be a global.

ltratt: > I think it makes sense to let true, false, nil be special. For system I'd think it should be a global.

Works for me! @ummarikar let's go with this.

lang_tests/system_global.som
```diff
@@ -0,0 +1,35 @@
...
+    String
+    21
+    a
+    b
+    c
+"
+
+system_global = (
+    run = (
+        system global: #system put: 45.
```
ummarikar: This line may seem out of place but since changing the `system` global in other SOMs produces a segmentation fault/ infinite loop then we know if we run this line and the rest of the test performs as planned then `system` is unchangeable (which is the behaviour we want in yksom).

src/lib/compiler/instrs.rs
```diff
@@ -12,18 +12,11 @@ pub enum Instr {
     Send(usize, usize),
     String(usize),
     Symbol(usize),
+    System,
```
ummarikar: Since there is only one "builtin" I removed the Builtin enum to replace is with this `System` instruction instead.

src/lib/compiler/ast_to_instrs.rs
smarr: @ummarikar I saw your last commit, and I'd prefer if `system` can be update as any other global. `true`, `false`, `nil` should be the ones that cannot be updated, or where the update has no effect.

smarr: I added a corresponding issue for a spec: https://github.com/SOM-st/SOM/issues/26

ummarikar: @smarr sorry! Misunderstood will send an update shortly!

smarr: No worries, I am also a little slow here with thinking about all the stuff related to globals. Turns out there's a little more :)

If a global is accessed before setting it, it raises an error by sending `#unknownGlobal:` with a symbol of the access global to `self`:

```
UnknownGlobal = (
  | excutedHandler ugName |
  unknownGlobal: name = (
    executedHandler := true.
    ugName := name.
    ^ 44
  )
  
  test = ( ^ nonExistingGlobal == 44 )
  verify = (
    ^ executedHandler and: [ ugName == nonExistingGlobal ]
  )
)
```

smarr: Ah, and `test` and `verify` are both expected to return `true`

ummarikar: Just so I understand. If I add this `unknownGlobal:`, then I can change what is returned when a user tries to reference a global that doesn't exist?

smarr: Yes, the `Object` class uses this to implement class loading:
https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L81

and here:
https://github.com/SOM-st/SOM/blob/master/Smalltalk/System.som#L53-L60

ummarikar:
Fixed the globals, now working on `unknownGlobal:`.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -485,13 +487,27 @@ impl<'a> Compiler<'a> {
...
+
+                        match lex_string {
+                            "nil" => vm.instrs_push(Instr::Builtin(Builtin::Nil)),
+                            "false" => vm.instrs_push(Instr::Builtin(Builtin::False)),
+                            "system" => {
+                                vm.instrs_push(Instr::Global(vm.add_symbol("system".to_string())))
+                            }
+                            "true" => vm.instrs_push(Instr::Builtin(Builtin::True)),
+                            _ => {
+                                if vm.is_global(lex_string) {
```
ummarikar: Hey @ltratt, @smarr mentioned here https://github.com/softdevteam/yksom/pull/90#discussion_r373776523 that I need to call `unknownGlobal:` on self. I am assuming I have to do that above. I made some initial attempts but they all failed (possibly due to my lack of understanding). Any chance you could point me in the right direction?

smarr: I am not sure I am reading the code correctly, though it looks to me as if you're doing the check at compile (AST to instruction) time.

The check needs to be performed at run time though.
So, in case of SOM (Java), I got a check as part of the `pushGlobal` instruction here:
https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Interpreter.java#L94-L104

ltratt: @ummarikar I think you might be in danger of trying to do too much in one go? Based on @smarr's comments, maybe an easier first PR would be to make `false`, `nil`, and `true` the only `Builtin`s and have every other free variable be handled a different way. That would require a very simple `system` class that, I think, can just do `get` (or equivalent). Once that's done then I think one can get into being clever with things like `unknownGlobal`.

ummarikar: Sounds good. I have made `false`, `nil` and `true` Builtins. I'll start working on a temporary method for handling free variables.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -485,13 +487,23 @@ impl<'a> Compiler<'a> {
...
+                            "nil" => vm.instrs_push(Instr::Builtin(Builtin::Nil)),
+                            "false" => vm.instrs_push(Instr::Builtin(Builtin::False)),
+                            "system" => {
+                                vm.instrs_push(Instr::Global(vm.add_symbol("system".to_string())))
+                            }
+                            "true" => vm.instrs_push(Instr::Builtin(Builtin::True)),
+                            _ => {
+                                vm.instrs_push(Instr::Global(
+                                    vm.add_symbol(lex_string.to_string()),
+                                ));
```
ummarikar: I realised that it should be possible to just push the `Global` instruction onto the vm's instructions stack so that any unknown free variables trigger the error here - https://github.com/softdevteam/yksom/blob/75b65d902ec7914b22a93ceb6484aedeb0c4c09a/src/lib/vm/core.rs#L363 but the test here - https://github.com/softdevteam/yksom/blob/34c71cdc81ddba6693089264bc7ea5ce60018c8d/lang_tests/instance_vars2.som#L11 returns an `UnknownMethod("run")` error. Why does this happen?

ummarikar: @smarr I am wondering what the https://github.com/softdevteam/yksom/blob/34c71cdc81ddba6693089264bc7ea5ce60018c8d/lang_tests/instance_vars2.som#L11 test should produce. I am getting the following error in Java SOM - `Runtime Error: ../Test.som:3:10: error: Could not parse double. Expected a number but got '^': 		x = (^x)`

ummarikar: For yksom, I have made it so that an `InvalidSymbol` error is returned, the same error returned when `system global: #ab` runs (if the global `#ab` has not yet been initialised).

ummarikar:
Assuming, @ltratt and @smarr, you are both happy with this, this could be a good place to end this initial PR.

src/lib/compiler/ast_to_instrs.rs
smarr: @ummarikar I don't see the parser error.

I get instead this:
```
/som.sh -cp Smalltalk ./instance_vars2.som
ERROR: Method run not found in class instance_vars2
```

This is expected, because, there's not `#run` method.
After adding one:

```
instance_vars2 = (
    x = (^x)
    run = (
      ^ self x
    )
)
```

I get:
```
ERROR: Unable to resolve x
```

I suppose as a first step before adding `#unknownGlobal`, your `InvalidSymbol` error makes sense.

For the previous question on `UnknownMethod("run")`: I don't know how you're executing the test.
A standard SOM expects a class that is given on the command line to implement a `#run` method similar to a `main` in C, Java, and other languages.
That's the `#run` method I added above.

lang_tests/system_global.som
```diff
@@ -0,0 +1,36 @@
...
+        Integer class println.
+        system global: #ab put: 21.
+        (system global: #ab) println.
+        ab println.
+        system global: #nil put: 'a'.
+        system global: #true put: 'b'.
+        system global: #false put: 'c'.
+        nil println.
+        true println.
+        false println.
```
smarr: Looks good to me.

src/lib/vm/core.rs
```diff
@@ -221,12 +229,33 @@ impl VM {
         Err(())
     }
 
+    /// Return whether the given string is a global.
+    pub fn is_global(&self, s: &str) -> bool {
```
ltratt: The comment says "is the given *string* a global" but uses `reverse_symbols`. I *think* the comment should say `Is the given symbol a global`? But then I'm not sure what type `s` should have. However, I'm not sure this function is even called anywhere?

src/lib/vm/core.rs
```diff
@@ -221,12 +229,33 @@ impl VM {
...
             .find_class(name)
             .unwrap_or_else(|_| panic!("Can't find builtin class '{}'", name));
-        self.compile(&path, inst_vars_allowed)
+
+        let val = self.compile(&path, inst_vars_allowed);
+        let idx = self.add_symbol(name.to_string());
+        let globals = unsafe { &mut *self.globals.get() };
+
+        if !globals.get(&idx).is_some() {
+            globals.insert(idx, Inst::new(self, val.clone()));
```
ltratt: This looks a bit fishy to me. As I read it it means "for every class C that we load, create a new object from it and place it into the globals"? I was expecting this to just be `globals.insert(val.clone())` I think.

src/lib/vm/core.rs
```diff
@@ -478,6 +516,37 @@ impl VM {
                 ));
                 SendReturn::Val
             }
+            Primitive::Global => {
+                let stack = unsafe { &mut *self.stack.get() };
```
ltratt: Simplifying a bit, with any of these `UnsafeCell`s, it's dangerous (i.e. we shouldn't do it) to assign them to a variable. Each time you use them, you should re-call `unsafe { ...get() }` to ensure that nothing bad happens *unless* you can prove in a comment that what you're doing is safe.

src/lib/compiler/ast_to_instrs.rs
ummarikar: Yes fixed in the new commit. Laurie pointed this out to me earlier not sure how I missed that the run method was missing probably because I am so used to it always being present! Thanks for the help!

src/lib/vm/core.rs
ummarikar: Since we are not doing an `is_global` check anymore I don't think we even need this function - removing it now.

src/lib/vm/core.rs
ummarikar: Agreed, not sure why that is there either!

ummarikar: Ahh I remember why that is there. https://github.com/softdevteam/yksom/blob/4e538bc719620c80655a10961d32b3ffefad1605/lang_tests/int_class.som#L10 This test fails otherwise (it returns `Class` instead of `Integer`). I think I made it an instance of a class to be the value the `global` is mapped to.

smarr: The test looks wrong. Class would be expected here. You ask the Integer class for its class, so it should return the meta class called 'Integer class':

```
(Integer class) println
Integer class
```

smarr:
```
---> Integer println
Integer
---> Integer class println
Integer class
---> Integer class class println
Metaclass
---> Integer class class class println
Metaclass class
---> Integer class class class class println
Metaclass
---> Integer class class class class class println
Metaclass class
```

smarr: But Metaclasses are probably also beyond this PR :)

src/lib/compiler/ast_to_instrs.rs
ltratt: > ERROR: Unable to resolve x

@smarr Just to check: this is a run-time error?

src/lib/vm/core.rs
ltratt: > But Metaclasses are probably also beyond this PR :)

I *think* (but I must admit that I've now forgotten) that yksom has support for meta-classes in this style. I think the `int_class` test probably wasn't tested on normal SOM @ummarikar?

src/lib/compiler/ast_to_instrs.rs
smarr: Yes, the set of globals is dynamic, for instance to support dynamic class loading. And to support arbitrary `system global: var put: val`

src/lib/vm/core.rs
ummarikar: It was tested but I naively assumed that “Integer” was yksoms version of “Integer class”. However, I was not able to find anything on metaclasses in yksom unless I am missing something...

src/lib/compiler/ast_to_instrs.rs
ltratt: Cool, I thought so, but wasn't sure.

@ummarikar I think that this PR means that it is no longer possible for there to statically be unknown variables which means that this part:

```rust
Err(vec![(*lexeme, format!("Unknown variable '{}'", name))])
```

in `find_var` can no longer be reached, right?

src/lib/vm/core.rs
ltratt: yksom doesn't have a `Metaclass` yet, but it has most of the machinery necessary to introduce this IIRC.

src/lib/compiler/ast_to_instrs.rs
ummarikar: It is “reached” but the error is caught. But, yes this error will never be seen.

ltratt: Then it should be removed in this PR if it can't be reached.

src/lib/vm/core.rs
ummarikar: Shall I remove this test and Metaclass can be done in a future PR?

ltratt: The test doesn't need `Metaclass`. The expected output should be `Class`. I think yksom does that, right?

ummarikar: SOM returns “Integer class”

ltratt: That looks to me simply like a minor difference in formatting output. You might want to see if it can easily be fixed or not.

ummarikar: Agreed, I’ll do that

smarr: I am not disagreeing with changing your test for now, but I would like to point out that it is a conceptual difference to have class objects that have a name field, and simply printing a generic string. SOM expects the former, and this likely requires metaclasses to get it right.

ltratt: @smarr Dumb question: where in SOM does the `<name> class` bit come from? I'm not sure where the "class" suffix is added, and a quick grep didn't make anything clearer.

smarr: In SOM-java, the class is assembled here: https://github.com/SOM-st/som-java/blob/master/src/som/compiler/ClassGenerationContext.java#L110-L112

Looking at SOM++, it seems to be in the same spot:
https://github.com/SOM-st/SOMpp/blob/master/src/compiler/ClassGenerationContext.cpp#L91-L93

Note that the methods always create two classes. The actual "result" class, as well as "the class" of the result class.

ltratt: Ah, I had been looking in the `.som` files, not the VM itself. Mea culpa! Thanks fo rthe link.

ummarikar: I have started getting this test to work unfortunately a temporary solution is still complicated as if we just change the `name` field in `Class` from `Integer` to `Integer class` it will break existing tests e.g. `1 class` will return `Integer class` instead of `Integer`.

smarr: Sounds to me like you got one class object too few :)

ummarikar: I think you might be right! Looking into meta classes now

ummarikar:
After jumping down a million rabbit holes, determined to figure it out myself, I have finally come up with something. At the vm level, `Class` and `Metaclass` use the same struct with the addition of `class`, which stores the class of a particular instance, and `metaclass`, which is a `bool`: true meaning the instance is a `Metaclass` and false meaning the instance is a `Class`.

ummarikar:
@ltratt I am not sure if what I have done is a hack or not but please let me know if there is anything I can do differently.

lang_tests/class_and_metaclass.som
```diff
@@ -0,0 +1,20 @@
...
+
+class_and_metaclass = (
+    run = (
+        Integer println.
+        (Integer class) println.
+        (Integer class class) println.
+        (Integer class class class) println.
+        (Integer class class class class) println.
+    )
+)
```
ummarikar: @smarr I ran this test in Java SOM and got the same output. Are you happy with this?

smarr: Looks fine to me. Thanks.

smarr:
I am not sure why the boolean is necessary. And I do not understand the logic where the boolean is tested.

In other SOM implementations, this boolean doesn't seem to be necessary. Metaclasses are also just instances of `Class`.

```
---> Integer class class superclass println
Class
```

Again, your PR is growing :) So, perhaps this is fine for the moment, but probably wants to be looked at at a later time again.

ltratt:
I agree with @smarr that we need to get this PR to a resolution. It's partly my fault for asking questions.

@ummarikar I suggest splitting this PR into two. Let's make this PR everything *before* the "adding metaclass" commit. You can do something like `git checkout -b metaclass ; git checkout class_primitivies; get reset --hard 4e538bc719` to do this. Then we can work out what we need to do to fix the pre-metaclass stuff. Maybe the simplest correct thing is to just delete that `int_class` test that was causing us confusion? Once we've got this PR to a point we're happy with, we can then move on to the actual metaclass PR.

ummarikar:
Done

lang_tests/system_global.som
```diff
@@ -0,0 +1,36 @@
...
+    42
+    String
+    21
+    21
+    nil
+    true
+    false
+"
+
+system_global = (
```
ltratt: To double check: does this test run identically on another SOM?

src/lib/vm/core.rs
```diff
@@ -226,7 +234,16 @@ impl VM {
         let path = self
             .find_class(name)
             .unwrap_or_else(|_| panic!("Can't find builtin class '{}'", name));
-        self.compile(&path, inst_vars_allowed)
+
+        let val = self.compile(&path, inst_vars_allowed);
+        let idx = self.add_symbol(name.to_string());
+        let globals = unsafe { &mut *self.globals.get() };
```
ltratt: As elsewhere, we can't (generally) safely assign `UnsafeCell`s to a variable, so unless we really need to, let's not do so.

In this case, we can probably change the code below to something like: `unsafe { &mut *self.globals.get() }.entry(&idx).and_modify(|e| *e = val)` (notice that I think the `Inst::new` is still bogus, right?).

src/lib/vm/core.rs
```diff
@@ -326,6 +343,15 @@ impl VM {
                     unsafe { &mut *self.stack.get() }.push(Double::new(self, i));
                     pc += 1;
                 }
+                Instr::Global(symbol_off) => {
+                    debug_assert!(unsafe { &*self.symbols.get() }.len() > symbol_off);
+                    if let Some(global) = unsafe { &mut *self.globals.get() }.get(&symbol_off) {
```
ltratt: We can probably do the `entry` trick here too.

ltratt: Ah, no, ignore that comment -- sorry!

src/lib/vm/core.rs
```diff
@@ -478,6 +504,31 @@ impl VM {
...
+            }
+            Primitive::GlobalPut => {
+                let value = unsafe { &mut *self.stack.get() }.pop();
+                let name = unsafe { &mut *self.stack.get() }.pop();
+                let as_string: &String_ = stry!(name.downcast(self));
+                let s = as_string.as_str();
+
+                if let Some(i) = unsafe { &mut *self.reverse_symbols.get() }.get(s) {
+                    unsafe { &mut *self.globals.get() }.insert(*i, value.clone());
+                    unsafe { &mut *self.stack.get() }.push(value.clone());
```
ltratt: My guess is that the second `clone` is unnecessary, but I might be wrong.

lang_tests/system_global.som
ummarikar: Everything except the last three lines but per @smarr 's request we kept `nil`, `true` and `false` as non-globals.

ltratt: Right, got it.

ummarikar:
Done.

src/lib/vm/core.rs
```diff
@@ -226,7 +234,14 @@ impl VM {
         let path = self
             .find_class(name)
             .unwrap_or_else(|_| panic!("Can't find builtin class '{}'", name));
-        self.compile(&path, inst_vars_allowed)
+
+        let val = self.compile(&path, inst_vars_allowed);
+        let idx = self.add_symbol(name.to_string());
+        unsafe { &mut *self.globals.get() }
+            .entry(idx)
+            .or_insert(val.clone());
```
ummarikar: `and_modify` did not pass the tests but this does.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -485,13 +487,23 @@ impl<'a> Compiler<'a> {
...
-                        "true" => vm.instrs_push(Instr::Builtin(Builtin::True)),
-                        _ => return Err(e),
-                    },
+                    Err(_) => {
+                        let lex_string = self.lexer.lexeme_str(&lexeme);
+
+                        match lex_string {
+                            "nil" => vm.instrs_push(Instr::Builtin(Builtin::Nil)),
+                            "false" => vm.instrs_push(Instr::Builtin(Builtin::False)),
+                            "system" => {
```
ltratt: Isn't this clause unnecessary in the sense that it's covered by the `_` clause below?

ummarikar: Yes, must have forgotten to remove it. Done now

ltratt:
Please squash.

src/lib/vm/core.rs
```diff
@@ -44,6 +44,8 @@ pub enum VMError {
         expected: ObjType,
         got: ObjType,
     },
+    /// Tried to access a global before it being initialised.
```
ummarikar: The comment that was here before was incorrect so I changed it to this. Everything is squashed now!

ltratt:
Should these 3 commits be squashed down into 1? The last commit seems to fix fairly big problems in the first 1 or 2 commits, so I'm not sure those are very useful to keep in the history. However, I can be convinced otherwise.

ummarikar:
You make a very good point I'll squash this into one.

ummarikar:
Squashed.

ltratt:
Would I be right in thinking that the commit title should really be "add support for globals"?

ummarikar:
Yes you would, one second.

ummarikar:
Done :)

PR #91, opened 2020-02-08
Add superclass primitive to `Class`.

ummarikar:
It should be possible to return the superclass of a given class in SOM.

### DIFF
--- a/src/lib/vm/objects/class.rs
+++ b/src/lib/vm/objects/class.rs
@@ -52,4 +52,11 @@ impl Class {
                 None => Err(Box::new(VMError::UnknownMethod(msg.to_owned()))),
             })
     }
+
+    pub fn superclass(&self, vm: &VM) -> Val {
+        if let Some(superclass) = &self.supercls {
+            return superclass.clone();
+        }
+        vm.nil.clone()
+    }
 }
--- a/lang_tests/class_superclass.som
+++ b/lang_tests/class_superclass.som
@@ -1,0 +1,18 @@
+"
+VM:
+  status: success
+  stdout:
+    Object
+    nil
+    String
+    nil
+"
+
+class_superclass = (
+    run = (
+        Integer superclass println.
+        Object superclass println.
+        Symbol superclass println.
+        Class superclass superclass println.
+    )
+)
--- a/lib/SOM/Class.som
+++ b/lib/SOM/Class.som
@@ -4,4 +4,6 @@ Class = (
     name     = primitive
 
     asString = ( ^self name )
+
+    superclass = primitive
 )
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -262,6 +262,7 @@ impl<'a> Compiler<'a> {
                 "printNewline" => Ok(MethodBody::Primitive(Primitive::PrintNewline)),
                 "printString:" => Ok(MethodBody::Primitive(Primitive::PrintString)),
                 "restart" => Ok(MethodBody::Primitive(Primitive::Restart)),
+                "superclass" => Ok(MethodBody::Primitive(Primitive::Superclass)),
                 "value" => Ok(MethodBody::Primitive(Primitive::Value(0))),
                 "value:" => Ok(MethodBody::Primitive(Primitive::Value(1))),
                 "value:with:" => Ok(MethodBody::Primitive(Primitive::Value(2))),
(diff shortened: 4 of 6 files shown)


ummarikar:
Tests pass and I tested this with Java SOM.

src/lib/vm/objects/class.rs
```diff
@@ -52,4 +52,11 @@ impl Class {
                 None => Err(Box::new(VMError::UnknownMethod(msg.to_owned()))),
             })
     }
+
+    pub fn superclass(&self, vm: &VM) -> Val {
+        if let Some(superclass) = &self.supercls {
```
ltratt: This can be simplified to something like `self.supercls.map(|x| x.clone())`.

smarr:
You may want to keep this PR as simple as it is, but just to move the discussion a little forward,
the notion of super class is again touching on the question of metaclasses:

```
---> 1 class println
Integer
---> 1 class superclass println
Object
---> 1 class superclass superclass println
nil
```

I think this works already as expected 👍 

But let's go into the `class` dimension as well:

---> 1 class class println
Integer class
---> 1 class class superclass println
Object class
---> 1 class class superclass superclass println
Class
---> 1 class class superclass superclass superclass println
Object
---> 1 class class superclass superclass superclass superclass println
nil
```

**Comment by ltratt** [comment]
@smarr Agreed, I think it's good not to tackle meta-classes in this PR, but useful to know what needs to be done!

@ummarikar However, @smarr makes a slightly separate good point: you should add a `class superclass superclass println` test I think.

**Comment by ummarikar** [comment]
> You may want to keep this PR as simple as it is, but just to move the discussion a little forward,
> the notion of super class is again touching on the question of metaclasses:
> 
> ```
> ---> 1 class println
> Integer
> ---> 1 class superclass println
> Object
> ---> 1 class superclass superclass println
> nil
> ```
> 
> I think this works already as expected 👍
> 
> But let's go into the `class` dimension as well:
> 
> ```
> ---> 1 class class println
> Integer class
> ---> 1 class class superclass println
> Object class
> ---> 1 class class superclass superclass println
> Class
> ---> 1 class class superclass superclass superclass println
> Object
> ---> 1 class class superclass superclass superclass superclass println
> nil
> ```

Yes, I got the idea for `superclass` from your comment in the previous PR I submitted. I will definitely include these tests when I submit the code for `metaclass` which is what I am eyeing for my next PR after this one.

**Comment by ummarikar** [comment]
@ltratt done and also tested with Java SOM to ensure I matched its output.

File: src/lib/vm/objects/class.rs [hunk 2]
```diff
@@ -54,8 +54,8 @@ impl Class {
     }
 
     pub fn superclass(&self, vm: &VM) -> Val {
-        if let Some(superclass) = &self.supercls {
```
ltratt: I was wrong: my suggestion has made the code worse! Let's revert this hunk.

ummarikar:
Done

ltratt:
Please squash.

ummarikar:
Squashed.

ltratt:
At first glance, that looks like a problem with rust-nightly to me. It might be worth waiting until there's a new build and trying again.

ltratt:
I think the new nightly is out (judging by https://rust-lang.github.io/rustup-components-history/), so let's try again.

bors r+

ltratt:
That still picked up yesterday's build. I'll try again later. @ummarikar In the meantime, just base a new PR off of this one and carry on. This should, in theory, be a temporary error.

ltratt:
So this is weird: if I use `rustup update` on a machine with an existing machine I get the 2002-02-05 version (which can compile yksom) but if I use the `rustup.sh` incantation in our `.buildbot.sh` file I get the 2002-02-09 version (which can't compile yksom). This feels pretty broken to me, but I haven't got a clue how to fix it, so I'm inclined to wait until later today and/or tomorrow and then try again.

ltratt:
If I install 2020-02-09 on my own user account on b13 things work. So either this is fixed, or there's something much weirder going on!

bors r+

ltratt:
I've now blown away all the `.cargo` directories in `buildbot-workerX` users, just in case.

bors r+

ltratt:
We believe we've fixed the issues in #93 -- if you rebase this PR against master, I can try bors again.

ummarikar:
Will do thanks for the help

ltratt:
Remember that you'll need to tell me if/when you've rebased.

ummarikar:
Yes sorry done

PR #94, opened 2020-02-14
Add `metaclass` to yksom.

ummarikar:
I added a boolean to `Class` to differentiate between `Class` and `Metaclass`. The reason why I did this is because `vm.meta_cls` is not yet initialised before compile-time and so must be returned at runtime instead. Also, I tried putting the `Class` object, that I return in `get_class` (https://github.com/softdevteam/yksom/blob/aaf937dfe3803c8e02d438daa293f845a72c924a/src/lib/vm/objects/class.rs#L36), inside the `Class` struct but this causes the superclass test to fail. This is the only implementation that I could get to pass all the tests and make (some) sense.

### DIFF
--- a/src/lib/vm/objects/class.rs
+++ b/src/lib/vm/objects/class.rs
@@ -7,12 +7,13 @@ use abgc_derive::GcLayout;
 
 use crate::vm::{
     core::{VMError, VM},
-    objects::{Method, Obj, ObjType, StaticObjType},
+    objects::{Method, Obj, ObjType, StaticObjType, String_},
     val::{NotUnboxable, Val},
 };
 
 #[derive(Debug, GcLayout)]
 pub struct Class {
+    pub metaclass: bool,
     pub name: Val,
     pub path: PathBuf,
     pub supercls: Option<Val>,
@@ -26,7 +27,23 @@ impl Obj for Class {
     }
 
     fn get_class(&self, vm: &VM) -> Val {
-        vm.cls_cls.clone()
+        let name: &String_ = self.name.downcast(vm).unwrap();
+        let s = name.as_str();
+
+        if !self.metaclass {
+            return Val::from_obj(
+                vm,
+                Class {
+                    metaclass: true,
+                    name: String_::new(vm, s.to_string() + " class", true),
+                    path: self.path.clone(),
+                    supercls: Some(vm.cls_cls.clone()),
+                    num_inst_vars: 0,
+                    methods: HashMap::with_capacity(0),
+                },
+            );
+        }
+        vm.meta_cls.clone()
     }
 }
 
--- a/src/lib/vm/objects/class.rs
+++ b/src/lib/vm/objects/class.rs
@@ -7,7 +7,7 @@ use abgc_derive::GcLayout;
 
 use crate::vm::{
     core::{VMError, VM},
-    objects::{Method, Obj, ObjType, StaticObjType, String_},
+    objects::{Method, Obj, ObjType, StaticObjType},
     val::{NotUnboxable, Val},
 };
 
--- a/src/lib/vm/objects/class.rs
+++ b/src/lib/vm/objects/class.rs
@@ -30,6 +30,11 @@ impl Obj for Class {
         let name: &String_ = self.name.downcast(vm).unwrap();
         let s = name.as_str();
 
+        let mut methods_clone = HashMap::with_capacity(self.methods.len());
+        for (k, v) in self.methods.iter() {
+            methods_clone.insert(k.clone(), Gc::clone(v));
+        }
+
         if !self.metaclass {
             return Val::from_obj(
                 vm,
@@ -39,7 +44,7 @@ impl Obj for Class {
                     path: self.path.clone(),
                     supercls: Some(vm.cls_cls.clone()),
                     num_inst_vars: self.num_inst_vars,
-                    methods: HashMap::clone(&self.methods),
+                    methods: methods_clone,
                 },
             );
         }
--- a/src/lib/vm/objects/class.rs
+++ b/src/lib/vm/objects/class.rs
@@ -13,7 +13,7 @@ use crate::vm::{
 
 #[derive(Debug, GcLayout)]
 pub struct Class {
-    pub metaclass: bool,
+    pub class: Option<Val>,
     pub name: Val,
     pub path: PathBuf,
     pub supercls: Option<Val>,
@@ -27,28 +27,10 @@ impl Obj for Class {
     }
 
     fn get_class(&self, vm: &VM) -> Val {
-        let name: &String_ = self.name.downcast(vm).unwrap();
-        let s = name.as_str();
-
-        let mut methods_clone = HashMap::with_capacity(self.methods.len());
-        for (k, v) in self.methods.iter() {
-            methods_clone.insert(k.clone(), Gc::clone(v));
+        match &self.class {
+            Some(cls) => cls.clone(),
+            None => vm.meta_cls.clone(),
         }
-
-        if !self.metaclass {
-            return Val::from_obj(
-                vm,
-                Class {
-                    metaclass: true,
-                    name: String_::new(vm, s.to_string() + " class", true),
-                    path: self.path.clone(),
-                    supercls: Some(vm.cls_cls.clone()),
-                    num_inst_vars: self.num_inst_vars,
-                    methods: methods_clone,
-                },
-            );
-        }
-        vm.meta_cls.clone()
     }
 }
 
@@ -66,6 +48,10 @@ impl Class {
     }
 
     pub fn get_method(&self, vm: &VM, msg: &str) -> Result<Gc<Method>, Box<VMError>> {
+        if self.class == None && self.supercls == None {
+            let cls: &Class = vm.cls_cls.downcast(vm).unwrap();
+            return cls.get_method(vm, msg);
+        }
         self.methods
             .get(msg)
             .map(|x| Ok(Gc::clone(x)))
@@ -78,6 +64,9 @@ impl Class {
     pub fn superclass(&self, vm: &VM) -> Val {
         if let Some(superclass) = &self.supercls {
             return superclass.clone();
+        }
+        if self.class == None {
+            return vm.cls_cls.clone();
         }
         vm.nil.clone()
     }
--- a/src/lib/vm/objects/class.rs
+++ b/src/lib/vm/objects/class.rs
@@ -38,8 +38,8 @@ impl Obj for Class {
                     name: String_::new(vm, s.to_string() + " class", true),
                     path: self.path.clone(),
                     supercls: Some(vm.cls_cls.clone()),
-                    num_inst_vars: 0,
-                    methods: HashMap::with_capacity(0),
+                    num_inst_vars: self.num_inst_vars,
+                    methods: HashMap::clone(&self.methods),
                 },
             );
         }
--- a/lang_tests/class_and_metaclass.som
+++ b/lang_tests/class_and_metaclass.som
@@ -7,6 +7,9 @@ VM:
     Metaclass
     Metaclass class
     Metaclass
+    Integer class
+    Metaclass
+    true
 "
 
 class_and_metaclass = (
@@ -16,5 +19,8 @@ class_and_metaclass = (
         Integer class class println.
         Integer class class class println.
         Integer class class class class println.
+        1 class class println.
+        1 class class class println.
+        (1 class <> (1 class class)) println.
     )
 ) 
--- a/lang_tests/class_and_metaclass.som
+++ b/lang_tests/class_and_metaclass.som
@@ -1,0 +1,20 @@
+"
+VM:
+  status: success
+  stdout:
+    Integer
+    Integer class
+    Metaclass
+    Metaclass class
+    Metaclass
+"
+
+class_and_metaclass = (
+    run = (
+        Integer println.
+        Integer class println.
+        Integer class class println.
+        Integer class class class println.
+        Integer class class class class println.
+    )
+) 
--- a/lang_tests/class_superclass.som
+++ b/lang_tests/class_superclass.som
@@ -8,6 +8,7 @@ VM:
     nil
     Class
     Class
+    Class
 "
 
 class_superclass = (
@@ -18,5 +19,6 @@ class_superclass = (
         Class superclass superclass println.
         Metaclass superclass println.
         Integer class class superclass println.
+        1 class class class superclass println.
     )
 )
--- a/lang_tests/class_superclass.som
+++ b/lang_tests/class_superclass.som
@@ -6,6 +6,8 @@ VM:
     nil
     String
     nil
+    Class
+    Class
 "
 
 class_superclass = (
@@ -14,5 +16,7 @@ class_superclass = (
         Object superclass println.
         Symbol superclass println.
         Class superclass superclass println.
+        Metaclass superclass println.
+        Integer class class superclass println.
     )
 )
--- a/lang_tests/class_superclass.som
+++ b/lang_tests/class_superclass.som
@@ -9,6 +9,7 @@ VM:
     Class
     Class
     Class
+    Class
 "
 
 class_superclass = (
@@ -20,5 +21,6 @@ class_superclass = (
         Metaclass superclass println.
         Integer class class superclass println.
         1 class class class superclass println.
+        Object class superclass println.
     )
 )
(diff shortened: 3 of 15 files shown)


ltratt:
How does Java SOM implement meta-classes? Does it use a boolean?

smarr:
SOMs normally support metaclasses by having separate instances of `Class` for them.

```
---> 1 class class println
Integer class
---> 1 class class class println
Metaclass
---> 1 class class class superclass println
Class
```

Note that these are proper objects, and operations such as identity need to work as expected.
And things like static fields and methods (or rather 'class-side' fields/method) need to work, too.

smarr:
Perhaps you want a test like this:

```
---> 1 class <> (1 class class)
it = true
```

ummarikar:
> SOMs normally support metaclasses by having separate instances of `Class` for them.
> 
> ```
> ---> 1 class class println
> Integer class
> ---> 1 class class class println
> Metaclass
> ---> 1 class class class superclass println
> Class
> ```
> 
> Note that these are proper objects, and operations such as identity need to work as expected.
> And things like static fields and methods (or rather 'class-side' fields/method) need to work, too.

My current implementation does differ wrt the use of the boolean, however, I still create a separate instance of `Class` for the `Metaclass` but it is created and returned at runtime. I have added some more tests now.

src/lib/vm/objects/class.rs
```diff
@@ -26,7 +27,23 @@ impl Obj for Class {
     }
 
     fn get_class(&self, vm: &VM) -> Val {
-        vm.cls_cls.clone()
+        let name: &String_ = self.name.downcast(vm).unwrap();
+        let s = name.as_str();
+
+        if !self.metaclass {
```
smarr: This looks strange.

Not entirely sure when this is triggered. Does `get_class` correspond to a `1 class` message?
Do you ensure `1 class class == (1 class class)`?

src/lib/vm/objects/class.rs
```diff
@@ -26,7 +27,23 @@ impl Obj for Class {
...
+            return Val::from_obj(
+                vm,
+                Class {
```
smarr: I read this as an instantiation of a new metaclass.
Normal SOMs do this when they instantiate the class object itself.

ummarikar:
@ltratt how would I go about doing this if `vm.meta_cls` is not initialised when the classes are created?

smarr:
Are you running into a bootstrapping problem?
Typically things have to be constructed step by step, and patched up on the way.

See
https://github.com/SOM-st/som-java/blob/master/src/som/vm/Universe.java#L331

and then https://github.com/SOM-st/som-java/blob/master/src/som/vm/Universe.java#L351
and later https://github.com/SOM-st/som-java/blob/master/src/som/vm/Universe.java#L380

ummarikar:
@ltratt Looking at what we have currently in yksom, it looks like we will have to rethink a lot in order to introduce metaclasses. I have tried to make some changes but I have been running into dead ends. I wonder if this should be my focus over the next couple of weeks when I should start writing dissertation soon. Or, if I should focus on somewhat smaller tasks and revisit this problem if it appears that I have time after writing my report. What do you think is best?

smarr:
@ummarikar Just asking since I do not know how yksom is structured. In some SOM VMs the class field is simply a mutable field like any other. How do you represent `myField` here, and can this technique be used for `class`, too?

Every object is supposed to have a class "field"* anyway.
So, for a `Class`, the `class` "field" simply contains the superclass.

```
MyClass = (
  | myField |
  set: val = ( myField := val. )
  get = ( ^ myField )
)
```

(*The field `class` is normally hidden from direct access, and access goes through a primitive.)

ummarikar:
> @ummarikar Just asking since I do not know how yksom is structured. In some SOM VMs the class field is simply a mutable field like any other. How do you represent `myField` here, and can this technique be used for `class`, too?
> 
> Every object is supposed to have a class "field"* anyway.
> So, for a `Class`, the `class` "field" simply contains the superclass.
> 
> ```
> MyClass = (
>   | myField |
>   set: val = ( myField := val. )
>   get = ( ^ myField )
> )
> ```
> 
> (*The field `class` is normally hidden from direct access, and access goes through a primitive.)

So currently at the rust-level there is no field in the `Obj` struct which stores `class` I have tried to implement some things that ended up being too hacky for my tastes and ultimately ended up breaking things (might be my lack of understanding of Rust, I am a Java person after all!). Forgive me if I have misunderstood, but are you saying I should add a variable `class` at the SOM level and setting this for every object?

smarr:
I had a brief look trying to find examples.

`Class` has a field for the superclass: https://github.com/softdevteam/yksom/blob/master/src/lib/vm/objects/class.rs#L18
It also needs one for the class.

I don't see a struct representing standard objects. I only see the build in objects.
The build in objects all have known classes, so, specializing them and simply having the `get_class` method return the right one seems fine.

However, for a normal object, I would expect somewhere a struct with a field storing the class.

> I should add a variable `class` at the SOM level and setting this for every object?

No, I would prefer that to happen at the Rust level.

ltratt:
> I don't see a struct representing standard objects. I only see the build in objects.
[...]
> However, for a normal object, I would expect somewhere a struct with a field storing the class.

I *think* what you're looking for can be found at 

However, for a normal object, I would expect somewhere a struct with a field storing the class.

smarr:
Hm, I think `Inst` and `Class` need to be some how "related". In class-based language, `Class` would be a subclass of `Inst`.

Note that a class can have fields, too.
A bunch of examples:

https://github.com/SOM-st/SOM/blob/0e89006bc4a3a8719a2ef841902d25bd72b7f43b/TestSuite/ClassA.som#L7

https://github.com/SOM-st/SOM/blob/0e89006bc4a3a8719a2ef841902d25bd72b7f43b/TestSuite/ClassC.som#L11

https://github.com/SOM-st/SOM/blob/0e89006bc4a3a8719a2ef841902d25bd72b7f43b/Examples/Benchmarks/JenkinsRandom.som#L3

https://github.com/SOM-st/SOM/blob/0e89006bc4a3a8719a2ef841902d25bd72b7f43b/Examples/Benchmarks/Json/JsonLiteral.som#L33

ltratt:
> Hm, I think Inst and Class need to be some how "related".

Yes, but:

>In class-based language, Class would be a subclass of Inst.

you can't naturally represent that relationship in Rust, so you need to find another way of encoding it. I'm not promising that yksom's encoding is the best one, but I *think* that extending it so that `Class`es can return `MetaClass`es should not be too awful. Of course, I haven't done the work, and anything involving bootstrapping can turn out to be awful for seemingly trivial reasons.

smarr:
A common Rust pattern seems to be a field in the struct with the "parent" struct. Don't know what you do in yksom.

In CSOM, we have a slightly terrible hack using macros.
In the other implementations, it's usually inheritance, since it's the "natural" way of doing it in those languages.

If one is careful enough, simply having the same fields in the struct per duplication also works of course. Might not be the worst engineering trade off, don't know.

ummarikar:
Wouldn't we need to use pointers as well to be able to do things like this - https://github.com/SOM-st/som-java/blob/3fab8ac7f34cedbbe15719fc607bceef25343d59/src/som/vm/Universe.java#L573?

ltratt:
My lazy assumption has long been that metaclasses probably require the strategic use of `UnsafeCell` somewhere (but, hopefully, only in one field). It might be that the right place is `Class::supercls` which perhaps should be `UnsafeCell<Option<Val>>` or similar. Warning: I might be *very* wrong on this!

ummarikar:
I had the same assumption. I was also thinking that I may need to do the same if we add a `class` field to the `Class` struct.

ummarikar:
I have been exploring "safe" options but I may try using `UnsafeCell` now.

ummarikar:
I am struggling to make any progress with metaclasses. Getting these pointers right in Rust seems to be really tricky with my limited knowledge of Rust and since I have spent over 10 days and have to stop coding soon to write my dissertation soon, I'll be moving on to working on relatively smaller tasks as I believe I could make the most of time left in this way. Closing this PR!

ummarikar:
Okay so I went down the `Array class` path as my next implementation and I realised that we need the metaclass in order to get this to work. This is because in yksom's current form `Array class println.` prints `Class` this is because we are returning `vm.cls_cls` whenever `class` is called on an object of type `Class`. This is a problem because it means we cannot get the methods that exist within the `Array` class object or any object of the `Class` type for that matter. `Class` needs to return a metaclass containing the methods that are available to that specific class. I hope I've explained this adequately.

Thus I have spent hours continuing to solve this metaclass issue. Now, if we are to mimic the same behaviour as Java - we need to initialise an instance of `Class` and then set the `class` field within it afterwards. So we need references. There are multiple ways to do that.

1. UnsafeCell - we cannot use this because we need to be able to mutate the underlying object in order to change the class.
2. Cell - Almost there, we can mutate the underlying data but our data needs to implement copy and Val cannot implement copy because it contains a destructor.
3. RefCell - This is most likely the path to take but I am currently getting already borrowed errors. Some really nasty stuff.

I am going to try and see if I can get RefCell to work.

ltratt:
I'm not sure I understand: what's the difference between, say, `String` and `Array` w.r.t. metaclasses?

smarr:
@ltratt I suspect the most important one is that `Array` has class-side methods: https://github.com/SOM-st/SOM/blob/master/Smalltalk/Array.som#L128-L155
While `String` doesn't have any.

ltratt:
Given that `Metaclass.som` is empty https://github.com/SOM-st/SOM/blob/master/Smalltalk/Metaclass.som, am I being an idiot if I think that one doesn't really *need* to support it in order for class-side methods to work?

ummarikar:
> Given that `Metaclass.som` is empty https://github.com/SOM-st/SOM/blob/master/Smalltalk/Metaclass.som, am I being an idiot if I think that one doesn't really _need_ to support it in order for class-side methods to work?

From what I understand whatever is returned by `get_class` needs to have a field `methods` which contains the methods that the SOM class has. We don't _need_ metaclasses in order to do that.

ltratt:
OK, so if we don't need metaclasses, why don't we side-step that problem entirely, assuming that your target is to implement `Array`?

ummarikar:
> OK, so if we don't need metaclasses, why don't we side-step that problem entirely, assuming that your target is to implement `Array`?

Sure I'll find a temporary solution for Array

smarr:
@ltratt If you want a uniform method lookup, as SOM has (no static methods like in Java), then you need an object. What's the type of the object representing the class?

ltratt:
@smarr I haven't really looked at SOM's semantics here: what I'm trying to work out is if there's a half-way reasonable hack that allows @ummarikar to keep going (seeing as metaclasses have proven a sizeable distraction at this point). Maybe there is no such hack!

smarr:
There's a hack: static method dispatch, a new type of dispatch.
But that's how you get Java.
And that's why we can't have nice things! :wink:

ummarikar:
If `abgc::Gc` implemented the `Clone` trait I can clone the `methods` hashmap in `class.rs` that way we construct a new `Class` struct every time the `get_class` method is called at runtime. This will solve the class method issue I am having while not breaking what we already have.

ltratt:
`Gc::clone(x)` should do the trick.

ummarikar:
`the trait bound abgc::Gc<vm::objects::method::Method>: std::clone::Clone is not satisfied`. This is the error I am getting when attempting the clone the hashmap.

ltratt:
`HashMap::clone` does not call `Gc::clone`. You'll need to manually clone the hashmap (e.g. using something like `iter().map(|x| Gc::clone(x))` if that's what you want to do.

ummarikar:
> `HashMap::clone` does not call `Gc::clone`. You'll need to manually clone the hashmap (e.g. using something like `iter().map(|x| Gc::clone(x))` if that's what you want to do.

Yeah thought so ... will do!

ummarikar:
UPDATE: I thought this might work but it doesn't I believe I am oversimplifying the problem somehow. The tests - symbol.som, class_and_metaclass.som, int29.som, class_superclass.som and system_global.som all fail. For context - system_global fails on line 25 it says `TypeError { expected: String_, got: Class }`. In class_and_metaclass the first line doesn't run it says that `to_strval` is not yet implemented. The other tests fail in similar ways, still trying to figure out the issue.

ummarikar:
All I did was clone the `methods` hashmap and include it in the class returned by `get_class`. This broke the above tests.

ltratt:
At first glance `system_global.som` looks broken to me. Line 25 is:

```
Integer class println.
```

but the test expects it to print `String`, which is surprising. So you may not have broken anything here: you may just have spotted a problem with the implementation of `system` or `Integer` or `Integer class` or similar.

ummarikar:
> At first glance `system_global.som` looks broken to me. Line 25 is:
> 
> ```
> Integer class println.
> ```
> 
> but the test expects it to print `String`, which is surprising. So you may not have broken anything here: you may just have spotted a problem with the implementation of `system` or `Integer` or `Integer class` or similar.

system `global` and `global:put:` looks fine. It looks like `Integer class println.` is the problem I wrote a test with just Integer class and it says `Exited due to signal: 9` and no extra information. What does this mean?

ltratt:
`Integer class println` should surely print `Integer` and not `String`? Given that that bug was introduced in a previous PR, I would fix that first, before trying to work out why you've got the "exited due to ..." error.

ummarikar:
It's not a bug from the previous PR, the class primitive doesn't cast anything to type `String_` so the `TypeError` can't be coming from there. It's something to do with the code I have submitted now.

ummarikar:
But, I'll have a look at the previous PR and see if there is anything that doesn't make sense.

ummarikar:
I just realised that because we are cloning `self.methods` we are forgetting about the `Class` methods that we should be able to perform as well. E.g. `Integer superclass println.`

ltratt:
I think the problem is present in master https://github.com/softdevteam/yksom/blob/master/lang_tests/system_global.som -- look at line 25 it's still `Integer class println` and its expected output is `String`.

ummarikar:
This is consistent with Java SOM. If you look at the line 24 we are changing the value returned by `Integer` to be the string `'a'`. We do this by `system global: #Integer put: 'a'`. So the expected output of `Integer class println.` should be `String`. This is not a bug.

ltratt:
Mea culpa! I understand now.

smarr:
> I just realised that because we are cloning self.methods we are forgetting about the Class methods that we should be able to perform as well. E.g. Integer superclass println.

The class does not have the same methods as the instance.
It has a different set of methods.

It only has the methods defined after `----` i.e., the switch to defining class-side methods.

```
---> 1 + 1
it = 2
---> Integer + 1
ERROR: Method + not found in class Integer class
```

Of course, there's still inheritance:

```
---> 1 class superclass println
Object
---> 1 class class superclass println
Object class
```

ummarikar:
I see that has cleared up some of my doubts thank you!

ltratt:
I've just checked and such methods aren't supported at all yet:

```
ClassMethods -> Result<(), ()>:
          "SEPARATOR" NameDefs MethodsOpt { unimplemented!() }
    | { Ok(()) }
    ;
```

where:

```
-----* "SEPARATOR"
```

so that suggests that quite a bit of the necessary machinery is missing.

ummarikar:
Okay, so I added support for class methods which turned out to fix every single problem we were having and subsequently gave me immense relief. I also implemented the `fromString:` class method for the `Integer` class and provided a test. This was a quick implementation just to show that I got class methods to work. It's been a painful two weeks but this seems to be a temporary hack that makes everything work!

ummarikar:
Added a test for the `ParseError`. I realise this PR is getting quite large but I saw the `fromString:` primitive as necessary in order to illustrate that class methods to indeed work now.

ltratt:
@ummarikar Because you need to be moving on to your report, I am going to suggest we leave this PR as-is for a while. We would need to split this PR up into bits and there's going to be a painful merge conflict with https://github.com/softdevteam/yksom/pull/95 if nothing else -- I don't think that is a useful way for you to spend your time at the moment. If you want to continue working on this, you should feel free to do so, and without having to worry about mergeability, at least not until after you've finished your report.

ummarikar:
Okay, that sounds like a good idea. I’ll start the report now and assuming I have time at the end I’ll revisit this PR and sort out the issues including the merge conflicts. Thanks!

ltratt:
Because this is now heavily behind master, I'm going to close it, as it will take quite a bit of work to bring it up to date. It can be reopened if you want.

PR #95, opened 2020-02-28
Move to using spans

ltratt:
The main purpose of this PR is to use `lrpar`'s `Span` type, which makes the grammar slightly uglier but the compiler slightly simpler -- since there's more complexity in the latter this seems like a reasonable trade.

### DIFF
--- a/src/lib/compiler/ast.rs
+++ b/src/lib/compiler/ast.rs
@@ -1,12 +1,10 @@
-use lrpar::Lexeme;
-
-type StorageT = u32;
+use lrpar::Span;
 
 #[derive(Debug)]
 pub struct Class {
-    pub name: Lexeme<StorageT>,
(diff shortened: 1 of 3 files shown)


ltratt:
bors try

ltratt:
@ptersilie See what you think about https://github.com/softdevteam/yksom/pull/95/commits/cb23546758596a555e90d4727183771a6961c993 -- I think it's a better name. If you agree, this PR will need squashing.

ltratt:
Because of https://github.com/softdevteam/yksom/pull/95/commits/9f7a9037c2909b7cac5b460215e75a50c6f0127a, this PR will definitely need squashing!

ptersilie:
Looks good to me. I agree that `CompileResult` is slightly better though `CompilerResult` would have also been ok. Please squash.

ltratt:
Squashed.

PR #96, opened 2020-03-01
block values need to execute with their self set to the outer object

ltratt:
The `value` primitive is rather special, and before this commit we didn't implement it quite correctly. The new test instance_var_in_block.som caused this error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
TypeError { expected: Inst, got: Block }', src/lib/vm/core.rs:353:39
```

The problem is that when executing a block via `value` we always set `rcv` to `Block`. This commit calls exec_primitive with an extra argument that allows it to set `rcv` to the "outer" `rcv`, fixing the problem.

### DIFF
--- a/lang_tests/instance_var_in_block.som
+++ b/lang_tests/instance_var_in_block.som
@@ -1,0 +1,15 @@
+"
+VM:
+  status: success
+  stdout:
+    1
+"
+
+instance_var_in_block = (
+    |y|
+    run = (
+        y := 0.
+        [ y := y + 1 ] value.
+        y println.
+    )
+)
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -95,7 +95,12 @@ impl<'a> Compiler<'a> {
                 .iter()
                 .map(|(lexeme, msg)| {
                     let ((line_off, col), _) = compiler.lexer.line_col(lexeme.span());
-                    let line = compiler.lexer.span_lines_str(lexeme.span()).split("\n").nth(0).unwrap();
+                    let line = compiler
+                        .lexer
+                        .span_lines_str(lexeme.span())
+                        .split("\n")
+                        .nth(0)
+                        .unwrap();
                     format!(
                         "File '{}', line {}, column {}:\n  {}\n{}",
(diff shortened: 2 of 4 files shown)


ltratt:
Let's hold off on this -- I may not have made the right fix.

ltratt:
I think that https://github.com/softdevteam/yksom/pull/96/commits/9384545f3f8d7847f288b57be6df3fa3ad5ff4bf is more likely to be correct (and, as a bonus, is much simpler). It'll take a bit longer for me to be sure, and this will certainly need squashing if it does turn out to be the right fix!

ummarikar:
This looks good to me assuming you're happy with it you can squash when you're ready

ltratt:
I'm now fairly convinced this is, at the very least, a substantial improvement (and maybe, who knows, even correct!). Squashed.

ltratt:
bors delegate=ummarikar

smarr:
I haven't looked at the unmodified yksom code, but does the following test work?


```
TestBlocks = (
  | field |
  method = ( ^ 5 )
  run = (
    | local |
    local := nil.
    [ | a |
      a := 100.
      a := a + [ | b |
        b := 1000.
        b := b + [ | c |
          c := 10000.
          c := c + [ | d |
            d := 100000.
            d := d + [ | e |
               e := 1000000.
               local := 30.
               e := e + self method + local.
               e ] value.
            d ] value.
          c ] value.
        b ] value.
      a ] value println.
    local println. 
  )
)
```

Expected output is:

```
1111135
30
```

ltratt:
Yep, it produces the expected output. Are you OK if I add this to the test suite or would you like to raise a PR?

[FWIW, I think that assigning to `self` in blocks might do surprising things in yksom, but I will look at that at a later date -- I'm currently doing some other work in the compiler, so would prefer not to make my merge hell worse than it is!]

smarr:
Sure, integrate it directly, if you think it's useful.

`self` should not be assignable: https://github.com/SOM-st/SOM/issues/6

ltratt:
> self should not be assignable: SOM-st/SOM#6

Ah, because Java SOM does allow it at the moment!

PR #98, opened 2020-03-04
Backtraces

ltratt:
Although SOM has the `halt` instruction to get a backtrace, I find it more useful to have a backtrace automatically printed when an error occurs. Thus yksom has had internal support from day one for percolating errors up the call stack.

This PR extends that such that we can print out detailed backtraces such as the following (for [this program](https://github.com/ltratt/yksom/blob/f76bff84c8abdb3890c365c9822ef63e24835996/lang_tests/nested_backtrace2.som)):

image

Note that this is not just printing out the call-stack, but it is highlighting the portion of line(s) which pertain to the error.

This PR is a bit scary, but there's a *lot* of boiler plate because we have to adjust lots of lang_tests. It's split into four bits:

 1) We track the `Span` that each instruction relates to https://github.com/softdevteam/yksom/commit/e222959159c91b7b34947bca4e37f7008da371ac
 2) We need to be able to track the call-stack with an error, so `VMError` needs to become a `struct` (the simple part of this in https://github.com/softdevteam/yksom/commit/f21e04090dd1471ea284e20fdcae76a3f6e36c86)
 3) Print out backtraces without highlighting (which requires changing lang_tests too) https://github.com/softdevteam/yksom/commit/4bcb29665486d00838fd0223290d5bb324412ed0
 4) Do the fancy terminal highlighting stuff https://github.com/softdevteam/yksom/commit/f76bff84c8abdb3890c365c9822ef63e24835996

Out of those it's really #3 that's the most imposing, and that's because of the lang_test adjustments which you can go over very quickly.

### DIFF
--- a/src/lib/vm/error.rs
+++ b/src/lib/vm/error.rs
@@ -1,0 +1,138 @@
+use std::fs::read_to_string;
+
+use abgc::Gc;
+use lrpar::Span;
+
+use crate::vm::{
+    core::VM,
+    objects::{Class, Method, ObjType},
+};
+
+#[derive(Debug)]
+pub struct VMError {
+    pub kind: VMErrorKind,
+    /// The callstack (in reverse order) of (Class, Span) pairs.
+    pub backtrace: Vec<(Gc<Method>, Span)>,
+}
+
+impl VMError {
+    pub fn new(vm: &VM, kind: VMErrorKind) -> Box<Self> {
+        let backtrace = Vec::with_capacity(vm.frames_len());
+        Box::new(VMError { kind, backtrace })
+    }
+
+    pub fn console_print(&self, vm: &VM) {
+        eprintln!("Traceback (most recent call at bottom):");
+        for (method, span) in self.backtrace.iter().rev() {
+            let cls_val = method.class();
+            let cls: &Class = cls_val.downcast(vm).unwrap();
+            let cls_path = cls.path.to_str().unwrap_or("<non-UTF8 filename>");
+
+            if let Ok(d) = read_to_string(&cls.path) {
+                let newlines = self.newlines(&d);
+                let (start_line, start_col) = self.line_col(&newlines, span.start());
+                let (end_line, _) = self.line_col(&newlines, span.end());
+                eprintln!(
+                    "  File {}, line {}, column {}:",
+                    cls_path, start_line, start_col
+                );
+                for i in start_line - 1..=end_line - 1 {
+                    let line = if i == newlines.len() {
+                        &d[newlines[i - 1]..]
+                    } else {
+                        &d[newlines[i - 1]..newlines[i]]
+                    };
+                    eprintln!("  {}", line.trim_end());
+                }
+            } else {
+                eprintln!("File {}:", cls_path);
+            }
+        }
+        eprintln!("{}.", self.kind.to_string());
+    }
+
+    fn newlines(&self, d: &str) -> Vec<usize> {
+        d.char_indices()
+            .filter(|(_, c)| *c == '\n')
+            .map(|(i, _)| i + 1)
+            .collect()
+    }
+
+    fn line_col(&self, newlines: &Vec<usize>, i: usize) -> (usize, usize) {
+        if newlines.is_empty() || i < newlines[0] {
+            return (1, i);
+        }
+
+        for j in 0..newlines.len() - 1 {
+            if newlines[j + 1] > i {
+                return (j + 2, i - newlines[j]);
+            }
+        }
+        (newlines.len() + 1, i - newlines[newlines.len() - 1])
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub enum VMErrorKind {
+    /// A value which can't be represented in an `f64`.
+    CantRepresentAsDouble,
+    /// A value which can't be represented in an `isize`.
+    CantRepresentAsIsize,
+    /// A value which can't be represented in an `usize`.
+    CantRepresentAsUsize,
+    DivisionByZero,
+    /// A value which is mathematically undefined.
+    DomainError,
+    /// The VM is trying to exit.
+    Exit,
+    /// Tried to access a global before it being initialised.
+    InvalidSymbol,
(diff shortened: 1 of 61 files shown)


smarr:
I think you'll find SOM implementations do have (hidden) support for stack traces:
- CSOM: https://github.com/SOM-st/CSOM/blob/master/src/vmobjects/VMFrame.c#L269
- SOM: https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Frame.java#L220

Though, they are typically not exposed directly, and often triggered from the debugger.

That's what `Object>>#halt` is for: https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L61

ltratt:
I stand corrected and have updated the PR description accordingly.

In fairness, there is a subtle difference here, which you can see (if you're prepared to wade through it) in the diff: yksom effectively implements catchable exceptions internally, but only catches them at the top level. Admittedly, the only reason I did that is because I want to know what effect this has on the style of implementation -- AFAIK SOM has no support for catching / retrying exceptions.

smarr:
Just for context: SOM tries to keep things simple, and as the PR shows, this is a great feature to have, but not necessarily a trivial extension.

And, well, then there's a cultural difference: as a Smalltalk, you are doing it wrong if you don't use a debugger! ;)

With respect to exceptions, yes, SOM doesn't have them.
Again, also just to keep the set of things minimal.
Non-local returns expose students to at least some of the aspects that are necessary to understand exceptions, and, exceptions are then a nice project for students to add.

ltratt:
FWIW, it was pretty trivial to implement this -- most of the churn in this PR is updating tests. I partly agree that there's a cultural difference about debuggers, although I still think most SOM implementations simply go "splat" when you do things like divide by zero, without giving you an opportunity to enter the debugger? Maybe I'm missing out on something obvious!

smarr:
> Maybe I'm missing out on something obvious!

Well, I run them in the debugger. I am using IDEs, but I suppose that translates to:

```
gdb --args ./CSOM  -cp Smalltalk
```

ltratt:
Ah, we're talking at different levels here. IMHO, unless I know the VM inside out (and often even if I know it inside out!), the VM-level backtrace isn't very helpful because it contains both too much detail (VM internal functions etc.) and too little (what source line number is this frame related to?): I need to debugging/backtracing at the source language level. I found this to be particularly the case with SOM, whose control-flow is quite different to what I'm used to.

smarr:
Well, typically I have stuff like this accumulating over time:

https://github.com/smarr/RoarVM/blob/master/vm/src/runtime/debug_helper.cpp

To be able to poke at things, print stack traces, etc.
For SOM, my Eclipse workspace has a whole bunch of frequently used snippets to poke at SOM/SOMns,...

Though, the fewest of these rather useful things get ever committed into repositories.

stephenrkell:
Enabling "source-, not VM-level" debugging in gdb is something I've been building up to for a while. Maybe it's time to try it in SOM?

(I like to think I pretty much know how it needs to work... although Actually Doing It has a habit of proving such beliefs wrong.)

src/lib/vm/objects/class.rs
```diff
@@ -15,6 +15,8 @@ use crate::vm::{
 pub struct Class {
     pub name: Val,
     pub path: PathBuf,
+    /// Where do this class's instructions start in VM::instrs?
```
ptersilie: Should we make this doc string a statement instead of a question? e.g. `Offset to this class's instructions in VM::instrs.`

ltratt: Fixed in https://github.com/softdevteam/yksom/pull/98/commits/f409739b239cd63bb58874e755994afa81537ce1.

src/lib/vm/error.rs
```diff
@@ -0,0 +1,141 @@
...
+                return (j + 2, i - newlines[j]);
+            }
+        }
+        (newlines.len() + 1, i - newlines[newlines.len() - 1])
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub enum VMErrorKind {
+    /// A value which can't be represented in an `isize`.
```
ptersilie: What's the difference between BigInt and isize? Both docstrings say "can't be represented as an isize".

src/lib/vm/error.rs
```diff
@@ -0,0 +1,141 @@
...
+    CantRepresentAsIsize,
+    /// A value which can't be represented in an `usize`.
+    CantRepresentAsUsize,
+    DivisionByZero,
+    /// A value which is mathematically undefined.
+    DomainError,
+    /// The VM is trying to exit.
+    Exit,
+    /// Tried to access a global before it being initialised.
```
ptersilie: Simplify to `Tried to access a global before it is initialised.`?

src/lib/vm/error.rs
ltratt: It turns out the first isn't used! I thought Rust would tell me about that, but it didn't. Fixed in https://github.com/softdevteam/yksom/pull/98/commits/3f3a096297cf55358a23e269d6fbdfc09589775e.

src/lib/vm/error.rs
ltratt: Ug, it turns out this is error is being used in two very different contexts. I think I'd better fix this one in another PR :/

src/lib/vm/error.rs
ptersilie: I gues it didn't tell you cause `VMErrorKind` is public, so it can't know.

src/lib/vm/error.rs
ptersilie: I was only referring to the docstring. Not sure what you mean.

ltratt: It turns out that the doc string is incorrect: this error is (incorrectly) used to signify two different classes of errors. I'd rather fix all that in a new PR.

ptersilie: Got it.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #99, opened 2020-03-06
Sort out globals

ltratt:
This is a fairly confusing part of SOM and we did not previously handle it properly. Assuming `unknown` is a variable not used in the program, then this code:

```
  unknown println.
```

is illegal but:

```
  global: #unknown
```

is legal (returning `nil`) and:

```
  unknown := ...
```

is illegal (complaining about "no field").

https://github.com/softdevteam/yksom/commit/bd8c47ba8b1e57d80da8c89a85f932401f5aacbd first fixes this.

https://github.com/softdevteam/yksom/commit/3d6f1a9e064018e12466830aea1ff3f2a05a3908 then tries to make yksom more like other SOMs in how it handles fiddling with "seemingly privilged" globals.


smarr:
Reading unknown globals should trigger `Object>>#unknownGlobal:`: https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L81

ltratt:
> Reading unknown globals should trigger Object>>#unknownGlobal:

I think I've tried all the tests in this PR on java-som, so I'm not sure if I don't have enough tests, or java-som or bugs -- or I managed not to test java-som properly!

smarr:
We had the very same discussion already. Reading unknown globals triggers the `#unknownGlobal:` handler, which in return tries to load a class.

The error message might be too concise, but says that it fails to resolve the class:

```
---> sss
ERROR: Unable to resolve sss
```

Error generated here: https://github.com/SOM-st/SOM/blob/master/Smalltalk/System.som#L71

ltratt:
Ah, yes, I see now. We don't have class loading machinery yet, so this is about as good as I can do at this moment in time. In the next PR I will probably add support for `unknownGlobal` though.

src/lib/vm/core.rs
```diff
@@ -475,28 +471,22 @@ impl VM {
...
-
-                if let Some(i) = unsafe { &mut *self.reverse_symbols.get() }.get(s) {
-                    if let Some(val) = unsafe { &mut *self.globals.get() }.get(&i) {
-                        unsafe { &mut *self.stack.get() }.push(val.clone());
-                        return SendReturn::Val;
-                    }
-                }
-                SendReturn::Err(VMError::new(self, VMErrorKind::InvalidSymbol))
+                let name_val = unsafe { &mut *self.stack.get() }.pop();
+                // XXX This should for Symbol.
```
ptersilie: There's something missing in this comment.

src/lib/vm/core.rs
```diff
@@ -475,28 +471,22 @@ impl VM {
...
-                let as_string: &String_ = stry!(name.downcast(self));
-                let s = as_string.as_str();
-
-                if let Some(i) = unsafe { &mut *self.reverse_symbols.get() }.get(s) {
-                    unsafe { &mut *self.globals.get() }.insert(*i, value.clone());
-                    unsafe { &mut *self.stack.get() }.push(value);
-                }
+                let v = unsafe { &mut *self.stack.get() }.pop();
+                let name_val = unsafe { &mut *self.stack.get() }.pop();
+                // XXX This should for Symbol.
```
ptersilie: Same as above.

src/lib/vm/core.rs
```diff
@@ -756,6 +746,75 @@ impl VM {
...
+        } else {
+            let globals = unsafe { &mut *self.globals.get() };
+            let len = globals.len();
+            reverse_globals.insert(s.clone(), len);
+            globals.push(Val::illegal());
+            len
+        }
+    }
+
+    /// Lookup the global `name`: if it has not been added, or has been added but not set, than
```
ptersilie: s/than/then

ptersilie:
Some typos.

ltratt:
Those two new commits should hopefully fix things.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #100, opened 2020-03-09
Class loading

ltratt:
This PR implements class loading, though it turns out we need to do a few bits first. The key commit is https://github.com/softdevteam/yksom/commit/d6561637a6c29ae2e3801662233bbe4e7b0c4aca. The mildly annoying bit before that is that the main interpreter loop was built with the implicit assumption that there was only one point where messages could be sent, whereas trying to read a value from an unassigned global calls `self unknownGlobal:`: https://github.com/softdevteam/yksom/commit/afd6fb86cc929e849d381ce9822fd811847ef1ec is the key commit enabling multiple message sending points in the main interpreter loop.


smarr:
Just to mentioned them, there are other places that may cause unexpected message sends:

- the `#perform*` methods https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L96
- any does-not-understand issue (but I suppose might still be in the loop, just when the original lookup failed) https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L72
- escaped blocks (activated after returned from stack) https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L77

ltratt:
Thanks @smarr!

ptersilie:
Looks good. Merge as is?

ltratt:
Yes please!

PR #102, opened 2020-03-11
Audit all uses of `UnsafeCell`s.

ltratt:
There were some incorrect uses of `UnsafeCell`, in particular (though not only) code such as the following:

```
  unsafe { &mut *self.stack.get() }.push(stry!(
    rcv.greater_than_equals(self, unsafe { &mut *self.stack.get() }.pop())
  ));
```

Notice that this holds two separate mutable references (without reborrowing) to the same `UnsafeCell` and is clearly incorrect. Rewriting it to:

```
  let v = stry!(rcv.greater_than_equals(self, unsafe { &mut *self.stack.get() }.pop()));
  unsafe { &mut *self.stack.get() }.push(v);
```

is both correct and, happily, easier to read!

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -316,14 +316,15 @@ impl VM {
                         let blkinfo = &unsafe { &*self.blockinfos.get() }[blkinfo_off];
                         (blkinfo.num_params, blkinfo.bytecode_end)
                     };
-                    unsafe { &mut *self.stack.get() }.push(Block::new(
+                    let blk = Block::new(
                         self,
                         Gc::clone(&method),
                         rcv.clone(),
                         blkinfo_off,
                         Gc::clone(&self.current_frame().closure),
                         num_params,
-                    ));
+                    );
+                    unsafe { &mut *self.stack.get() }.push(blk);
                     pc = bytecode_end;
                 }
                 Instr::ClosureReturn(closure_depth) => {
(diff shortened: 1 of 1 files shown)


jacob-hughes:
> Notice that this holds two separate mutable references (without reborrowing) to the same UnsafeCell and is clearly incorrect

While I think this is highly dangerous, I don't think that on it's own this is incorrect. Reborrowing in the Rust abstract machine is implicit with the second `&mut`. It certainly has the potential to be incorrect: any future re-use of the first borrow will invalidate the second one. 

I agree though that the second example is far easier to read.

ltratt:
It's definitely incorrect because it doesn't reborrow a reference -- it creates two separate references. Here's a Playground example showing how this can go wrong https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=63626668bdacb411f2e03de542f8165e -- run it under Miri and watch it bork.

jacob-hughes:
> because it doesn't reborrow a reference -- it creates two separate references

Aha! You're right. Wow that's subtle. It's UB when trying to create the second mutable reference because it cannot find an existing borrow in the borrow stack to re-borrow from.

ltratt:
> It's UB when trying to create the second mutable reference because it cannot find an existing borrow in the borrow stack to re-borrow from.

That's not how I think of it, because clearly there *is* an existing borrow in the borrow stack the second `unsafe` block could borrow from. The issue AFAICS is that you have to *explicitly* state which borrow in the the borrow stack you're reborrowing from, and this code doesn't: two references are created from the same unsafe "source", so to speak.

jacob-hughes:
I think you're right.

ltratt:
OK, if you're happy with the PR overall, you can r+ this, or if you have specific comments, I'll do my best to address them!

jacob-hughes:
Ok, I think I'm happy with the rest of the code changes. bors r+

ltratt:
I'm not sure, but I think that the bors command needs to come on its own line?

PR #103, opened 2020-03-12
Reduce the quantity of explicitly unsafe code

ltratt:
This PR is in two stages: first of all it gives `SOMStack` (the SOM stack of values) an immutable API (https://github.com/softdevteam/yksom/commit/d8cb3ce47c2495e590833efced89228f7bf55000) and then uses that to reduce the `unsafe` code in the main interpreter (https://github.com/softdevteam/yksom/commit/af582fe8138fc074c043e897ff8ff5fb944ac23e). This should make it harder to reintroduce the incorrect code we addressed in #102.

[Note that `SOMStack` is not thread-safe, but that's OK a) because the use of `Cell` means that it is not `Sync` b) SOM is single-threaded.]

### DIFF
--- a/src/lib/mod.rs
+++ b/src/lib/mod.rs
@@ -12,6 +12,7 @@
 
 #![feature(alloc_layout_extra)]
 #![feature(allocator_api)]
+#![feature(cell_update)]
 #![feature(coerce_unsized)]
 #![feature(specialization)]
 #![feature(unsize)]
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -78,7 +78,7 @@ pub struct VM {
     /// reverse_sends is an optimisation allowing us to reuse sends: it maps a send `(String,
     /// usize)` to a `usize` where the latter represents the index of the send in `sends`.
     reverse_sends: UnsafeCell<HashMap<(String, usize), usize>>,
-    stack: UnsafeCell<SOMStack>,
+    stack: SOMStack,
     strings: UnsafeCell<Vec<Val>>,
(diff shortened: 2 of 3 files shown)


smarr:
Having close to zero experience with Rust, I am a bit surprised that you would hide something as important as a side effect/mutation from the stack interface.

Is this because it is trivial and not necessary?
Stack discipline seems like something important. Though not sure whether the types help with it or not.

jacob-hughes:
@smarr 

> Is this because it is trivial and not necessary?

Without speaking for Laurie, :wink: I think it's more that by holding a mutable reference to `SOMStack`, a consumer of the API can unwittingly mutate fields it _shouldn't_ have been allowed to. Since in Rust it's not possible to have a Struct with some fields which are immutable and some which aren't, a common pattern is to use Interior Mutability: hold an immutable reference to the Struct, but "opt-in" to have some fields mutable through an opaque `Cell` like interface.

smarr:
Hm, but shouldn't the user have to do something, in order to be made aware of the side effect, when triggering pop/push? To the untrained eye, it looks like pop/push are now side effect free?

ltratt:
For better or worse, Rust APIs are often quite different than in other languages -- and the trade-offs aren't obvious at first. Since I started yksom I've certainly learnt a lot! At least in multi-threaded VMs, *somewhere* one will have to do without exterior mutability: the question is "where?". I've tried modelling that in yksom though I don't think I've got it right yet: I think I might have a nicer model that might work better for multi-threaded VMs. It's next on my list of things to try, but it might fail.

jacob-hughes:
This LGTM. Let me know if you're happy to bors it.

ltratt:
Please go ahead!

PR #104, opened 2020-03-13
Remove as many UnsafeCells as possible

ltratt:
This is a fairly long, but surprisingly mostly-not-too-difficult, PR that removes `UnsafeCell`s from nearly all parts of yksom. https://github.com/softdevteam/yksom/commit/4e13c53bcf5db87d2c24dc160344788f8ecb0aab is the biggie: it makes the core `VM` struct require `&mut self` which ripples throughout the entire system (e.g. requiring a fairly large number of new intermediate variables to keep the borrow checker happy). That commit also contains the explanation which, IMHO, justifies the rest of the PR. Overall, I think this is *probably* a good idea, because it's very clear that it's hard to use large numbers of `UnsafeCell`s correctly, but there is always the possibility that this makes the VM unergonomic in the long term (I suspect that https://github.com/softdevteam/yksom/commit/b2b9e26cbf3e9fdc922f0814c58f4e5afcaffc11 is probably the main way of making things unergonomic in the long term).

After the first commit, it's a mostly mechanical job to remove most other `UnsafeCell`s, with the notable exceptions of https://github.com/softdevteam/yksom/commit/b14d07dd8e209c36f98ecd5707aedd88a6fa01d2 and https://github.com/softdevteam/yksom/commit/09e37e2963cd363dd3e6a2bef92a28ecee43aa40.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -481,12 +481,14 @@ impl VM {
 
         match prim {
             Primitive::Add => {
-                let v = stry!(rcv.add(self, self.stack.pop()));
+                let v = self.stack.pop();
+                let v = stry!(rcv.add(self, v));
                 self.stack.push(v);
                 SendReturn::Val
             }
             Primitive::And => {
-                let v = stry!(rcv.and(self, self.stack.pop()));
+                let v = self.stack.pop();
+                let v = stry!(rcv.and(self, v));
                 self.stack.push(v);
                 SendReturn::Val
             }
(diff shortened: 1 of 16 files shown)


ltratt:
bors try

src/lib/vm/core.rs
```diff
@@ -324,7 +324,7 @@ impl VM {
             match instr {
                 Instr::Block(blkinfo_off) => {
                     let (num_params, bytecode_end) = {
-                        let blkinfo = &unsafe { &*self.blockinfos.get() }[blkinfo_off];
+                        let blkinfo = &self.blockinfos[blkinfo_off];
```
jacob-hughes: I much prefer this. It makes reasoning about what's happening a lot easier.

jacob-hughes:
This LGTM. I'm definitely all for removing `UnsafeCell`. I can't see anything wrong with it but there were quite a lot of changes. 

Will bors it if you're happy.

ltratt:
If you're happy, this can be bors'd.

PR #105, opened 2020-03-16
Class methods and fields

ltratt:
This PR adds support for class methods and fields, which also requires metaclasses too. I needed a bit of help from @smarr to understand this, because in essence if you define a class:

```
  C = (
    |x|
    m1 = ( 'm1' println. )
    ------
    |y|
    m2 = ( 'm2' println. )
  )
```

you're actually defining two classes. The class C is:

```
  C = (
    |x|
    m1 = ( 'm1' println. )
  )
```

and its metaclass:

```
  Cclass = (
    |y|
    m2 = ( 'm2' println. )
  )
```

The [`class_methods_fields.som`](https://github.com/ltratt/yksom/blob/f47e711582b4d6e9d43c1337904df82696445fcd/lang_tests/class_methods_fields.som) lang_test shows this in action. However, the first major step towards this is metaclasses (https://github.com/softdevteam/yksom/commit/31555a97c348860551892a83af057a4072bcd178), which is much trickier, as we have to deal with the fact that during VM initialisation we want to create references to objects which don't yet exist. Once that's done properly, most of the other bits fall into place relatively easily (famous last words).

@smarr Are there any tests you think should be added here?

### DIFF
--- a/lang_tests/class_methods_fields.som
+++ b/lang_tests/class_methods_fields.som
@@ -1,0 +1,21 @@
+"
+VM:
+  status: success
+  stdout:
+    #class_methods_fields
+    b
+"
+
+class_methods_fields = (
+    run = (
+        class_methods_fields name println.
+        class_methods_fields set: 'b'.
+        class_methods_fields get println.
+    )
+
+    ------
+    | f |
+
+    get = ( ^f )
+    set: o = ( f := o. )
+)
--- a/lang_tests/metaclasses.som
+++ b/lang_tests/metaclasses.som
@@ -1,0 +1,52 @@
+"
+VM:
+  status: success
+  stdout:
+    1
+    Integer
+    Integer class
+    Metaclass
+    Metaclass class
+    Metaclass
+    Metaclass class
+    Class
+    Object class
+    Class
+    Boolean class
+    Class
+    Object
+    nil
+    Class class
+    Metaclass
+    Object
+    Object class
+    Metaclass
+    Class
+    nil
+"
+
+metaclasses = (
+    run = (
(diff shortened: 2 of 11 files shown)


ltratt:
bors try

smarr:
> Are there any tests you think should be added here?
Ehm, good question.

Perhaps `Object superclass println` -> `nil`.

One might perhaps also test for the number of fields the class/classClass have.
https://github.com/SOM-st/SOM/blob/master/TestSuite/ClassStructureTest.som has a few tests around that.
- testAccessToInstanceFields
- testAccessToClassFields
- testInstanceFields

ltratt:
> Perhaps Object superclass println -> nil.

Good idea -- added with thanks!

Unfortunately I don't have some of the necessary machinery for the other tests you suggested yet :/ Maybe I should build some of that up in the next PR: I don't think we're missing *much* of it.

ptersilie:
`leaves `Val::ilegal`s lurking around, which, if dereferenced, cause bad things`

Two typos in commit message 31555a9: s/ilegal/illegal and s/cause/causes

ltratt:
`ilegal` is an embarrassing typo; but "cause" is correct grammar in this case IMHO.

ptersilie:
Shouldn't it be the same as `leaves`: `That leaves ... and causes ...`?

ptersilie:
Ah, sorry, didn't see the `s` after `Val::illegal`, making it plural. This makes more sense now.

ptersilie:
Ok, please squash.

ltratt:
Squashed (and reworded).

PR #106, opened 2020-03-18
Add stubs for more SOM primitives.

ltratt:
This gets us closer to being able to parse the SOM standard library.


smarr:
Just a note on primitives and modularity: 

There was always the idea that SOM has support for loading primitives dynamically.
CSOM and SOM++ support it.

One example is here: https://github.com/SOM-st/SOM/tree/master/Examples/Snake

(see the .c/.cpp files)

And then there are two other things that might be worthwhile to mention:
 - primitives can be either on the class or instance side (we have both variants in use)
 - primitives are installed on a specific class, which means, one can override them, and another class can have another primitive with the same name. I am not sure this is really used in practice. Though, one can technically use it, for example to optimize the number operations for the specific receiver class at hand.

ltratt:
I'm not going to try supporting dynamic linking -- that's too much work for too little benefit.

The way I've implemented primitives is currently very crude, but means that that primitives can naturally do the "right" thing, whether they're on classes or instances, and/or when they're overloaded. That said, I probably will soon make them a little less crude, because I can shift some checks off to compile-time, whereas they're all done at run-time at the moment. When I do that, I'll make sure to keep in mind the flexibility you've mentioned -- thanks for the pointers!

smarr:
Yeah, well, it was an "idea". Don't think it is essential.

I am just surprised that `ast_to_instrs.rs` and `instrs.rs` have anything to do with it.
Might not be ideal, especially if I comes along and randomly add new primitives.

ltratt:
You can add primitives, it's just done in a different way to other SOMs. AFAICS the two approaches are entirely transparent from the end user's perspective (modulo dynamic linking, which would introduce a world of pain that I'm not going to explore!). In other words, I think the differences are unimportant. I might be wrong!

PR #107, opened 2020-03-20
Switch out abgc for rboehm

jacob-hughes:
All tests for this should be pass, but let me know if you find bugs or have problems building things on OpenBSD!

### DIFF
--- a/lang_tests/system_global_lookup_string.som
+++ b/lang_tests/system_global_lookup_string.som
@@ -6,8 +6,10 @@ VM:
     xxx
 "
 
+
 system_global_unset_global = (
     run = (
         (system global: 'ab') println.
     )
 )
+
--- a/src/lib/vm/val.rs
+++ b/src/lib/vm/val.rs
@@ -6,10 +6,9 @@ use std::{
     convert::TryFrom,
     mem::{size_of, transmute},
     ops::Deref,
-    ptr::NonNull,
 };
 
-use abgc::{self, Gc};
+use rboehm::{self, Gc};
 use num_bigint::BigInt;
 use num_enum::{IntoPrimitive, UnsafeFromPrimitive};
 use num_traits::{FromPrimitive, ToPrimitive, Zero};
@@ -58,7 +57,7 @@ pub trait NotUnboxable {}
 
 /// The core struct representing values in the language runtime: boxed and unboxed values are
 /// hidden behind this, such that they can be treated in exactly the same way.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct Val {
     // We use this usize for pointer tagging. Needless to say, this is highly dangerous, and needs
     // several parts of the code to cooperate in order to be correct.
@@ -72,10 +71,10 @@ impl Val {
     /// `Obj` couldn't be a trait object. Oh well.]
     pub fn from_obj<T: Obj + 'static>(_: &mut VM, obj: T) -> Self {
         debug_assert_eq!(size_of::<*const ThinObj>(), size_of::<usize>());
-        let ptr = ThinObj::new(obj).into_raw();
+        let ptr = Gc::into_raw(ThinObj::new(obj));
         Val {
             val: unsafe {
-                transmute::<*const ThinObj, usize>(ptr.as_ptr()) | (ValKind::GCBOX as usize)
+                transmute::<*const ThinObj, usize>(ptr) | (ValKind::GCBOX as usize)
             },
         }
     }
@@ -93,9 +92,9 @@ impl Val {
     /// undefined behaviour.
     pub fn recover(obj: &dyn Obj) -> Self {
         unsafe {
-            let ptr = ThinObj::recover(obj).into_raw();
+            let ptr = Gc::into_raw(ThinObj::recover(obj));
             Val {
-                val: transmute::<*const ThinObj, usize>(ptr.as_ptr()) | (ValKind::GCBOX as usize),
+                val: transmute::<*const ThinObj, usize>(ptr) | (ValKind::GCBOX as usize),
             }
         }
     }
@@ -172,9 +171,9 @@ impl Val {
             ValKind::GCBOX => {
                 debug_assert_eq!(size_of::<*const ThinObj>(), size_of::<usize>());
                 Ok(unsafe {
-                    Gc::clone_from_raw(NonNull::new_unchecked(
+                    Gc::from_raw(
                         self.val_to_tobj() as *const _ as *mut _
-                    ))
+                    )
                 })
             }
             ValKind::INT => {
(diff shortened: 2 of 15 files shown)


lang_tests/system_global_lookup_string.som
```diff
@@ -6,8 +6,10 @@ VM:
     xxx
 "
 
+"
```
jacob-hughes: Oops. Not sure why I committed this.

ltratt: Any reason for commenting this out?

src/lib/vm/val.rs
```diff
@@ -543,40 +543,40 @@ binop_typeerror!(greater_than_equals, >=);
...
-                        self.val_to_tobj() as *const _ as *mut _
-                    ))
-                });
-            }
-            ValKind::INT => (),
-            ValKind::ILLEGAL => (),
-        }
-    }
-}
+// impl Clone for Val {
```
ltratt: I'm fine with deleting the `Clone` block.

src/lib/vm/val.rs
```diff
@@ -543,40 +543,40 @@ binop_typeerror!(greater_than_equals, >=);
...
+//                 ) | (ValKind::GCBOX as usize)
+//             },
+//             ValKind::INT => self.val,
+//             ValKind::ILLEGAL => self.val,
+//         };
+//         Val { val }
+//     }
+// }
+
+// impl Drop for Val {
```
ltratt: And I think we can delete the `Drop` block too.

lang_tests/system_global_lookup_string.som
jacob-hughes: I had problems running `ignored` tests locally in `lang_tester`,  I checked it in by mistake.

ltratt:
At face value, this looks pretty good: I can't build on OpenBSD due to an rboehm problem, but I'll look into that soon.

jacob-hughes:
What's the error, I might know it?

ltratt:
```
error: failed to run custom build command for `rboehm v0.1.0 (https://github.com/softdevteam/rboehm#06abbf6d)`

Caused by:
  process didn't exit successfully: `/home/ltratt/scratch/softdev/yksom/target/debug/build/rboehm-48afd70ca21c1cd7/build-script-build` (exit code: 101)
--- stderr
Cloning into '/home/ltratt/scratch/softdev/yksom/target/debug/build/rboehm-f4ab0a08ad9dc7a1/out/bdwgc'...
Cloning into 'libatomic_ops'...
/usr/local/bin/aclocal[35]: /usr/local/bin/aclocal-xfsettingsd.core: not found
autoreconf-2.69: aclocal failed with exit status: 127
thread 'main' panicked at 'failed to execute "./configure"', /home/ltratt/.cargo/git/checkouts/rboehm-e46be1172403150c/06abbf6/build.rs:20:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jacob-hughes:
Ah okay. I think you might be able to get away without building the `libatomic_ops` sub-library for now as we don't use any of the multi-threaded features of boehm

src/lib/vm/val.rs
```diff
@@ -9,7 +9,7 @@ use std::{
     ptr::NonNull,
```
ltratt: This seems unused now, so can be removed?

ltratt:
I realised that I should just benchmark this, just to see, even though we only have one (not very representative) benchmark: a while loop that counts down from *n* to 0. Before this PR:

```
===> multitime results
1: -q target/release/yksom --cp lib/SOM /home/ltratt/yksom/while_bench.som
            Mean                Std.Dev.    Min         Median      Max
real        0.396+/-0.0018      0.004       0.389       0.396       0.403
user        0.395+/-0.0021      0.004       0.385       0.394       0.403
sys         0.001+/-0.0007      0.001       0.000       0.000       0.004
```

This PR:

```
1: -q target/release/yksom --cp lib/SOM /home/ltratt/yksom/while_bench.som
            Mean                Std.Dev.    Min         Median      Max
real        3.266+/-0.0171      0.036       3.177       3.262       3.343
user        6.273+/-0.1033      0.220       5.959       6.225       6.625
sys         0.514+/-0.0311      0.066       0.372       0.522       0.675
```

In other words, for at least that one benchmark, this PR makes a meaningful improvement to performance, which isn't necessarily what I expected, but is very welcome!

smarr:
How do I read the numbers? Going from 0.4seconds of real time to 3.3seconds doesn't seem like an improvement.

ltratt:
I'm an idiot. Let me update that.

jacob-hughes:
I'm fairly sure a huge chunk of this slowdown is to do with our naive finalization approach. Here are some numbers I ran on a similar, countdown-from-_n_ benchmark:

With abgc:
```
===> multitime results: abgc
1. ./abgc/release/yksom --cp lib/SOM/ while_benchmark.som
            Mean        Std.Dev.    Min         Median      Max
real        0.719       0.040       0.663       0.733       0.775       
user        0.711       0.036       0.658       0.723       0.752       
sys         0.005       0.005       0.000       0.005       0.016       
```

With rboehm (this PR):
```
===> multitime results: abgc
1: ./target/release/yksom --cp lib/SOM/ while_benchmark.som
            Mean        Std.Dev.    Min         Median      Max
real        5.845       0.343       5.605       5.723       6.849       
user        9.244       0.799       8.726       9.009       11.599      
sys         0.377       0.043       0.331       0.365       0.493  
```

With rboehm, disabling finalizers:
```
===> multitime results: No Finalizers
1: ./no_finalizer/release/yksom --cp lib/SOM/ while_benchmark.som
            Mean        Std.Dev.    Min         Median      Max
real        1.572       0.047       1.506       1.572       1.653       
user        2.931       0.064       2.834       2.931       3.058       
sys         0.107       0.015       0.089       0.106       0.128       
```

By switching off finalizers, we go down from a slowdown of 8x to 2x when compared with abgc. Obviously, we can't just switch them off and be done with it, as we'll be leaking memory on any non-trivial data structure.

Unfortunately, only adding objects to a finalization queue when absolutely necessary requires compiler support. It is not as simple as looking for a `Drop` impl; we must hook into dropck to determine whether field destructors need running. However, there are some clear wins to be had in the meantime. I have an upcoming PR for rboehm which can prevent enqueuing objects for finalization when it can be certain it is unnecessary.

smarr:
@jacob-hughes what exactly does need "finalization" here? SOM does not support any notion of "external resources", which is what I associate finalization with. Or asked differently, what do you mean with finalization?

I am also curious, what's the motivation to switch to Boehm?
I am not sure what abgc is, is it an actual GC?

jacob-hughes:
@smarr abgc is a simplified reference counting library which yksom uses to manage objects. At the moment, Rust has no support for tracing GC (but I'm working on it!). A large part of this is getting the API right: having GC'd objects interoperate smoothly with normal RAII Rust values can be particularly tricky. Boehm makes sense as a suitable backend for testing this since it's very flexible to use and right now we're not focused on the collection algorithm.

> SOM does not support any notion of "external resources", which is what I associate finalization with.

Finalization in this sense is orthogonal to SOM. Rust uses a `Drop` trait to implicitly run an object's destructor once it goes out of scope (somewhat analogous to C++). We allow GC things to point to non-GC things and vice versa. If a non-GC object with a destructor is owned by a GC object, that destructor will need running when the GC object is reclaimed.

The non-deterministic nature of GC throws a spanner in the works here: Rust can't automatically call an object's `drop` destructor because it won't know when it's dead. Instead, the collector must trigger the destructors on non-GC'd objects when the GC object which owns them dies. We do this by using finalizers which explicitly call inner destructors. This is a bit of an anti-pattern, because finalizers are expected to be rare. 

As for my previous comment: right now we err on the side of caution and register finalizers to invoke inner destructors "just in case". There are optimisations which prevent this being necessary which will be easy wins.

ltratt:
@jacob-hughes Thanks for that insight, it's very useful and hopeful!

Can I suggest you add the change required by https://github.com/softdevteam/rboehm/pull/5 in to this PR, then we can merge?

src/lib/vm/val.rs
jacob-hughes: Fixed: 011d3fd58a949c271c13f385e2f9a6cc107e1a99

src/lib/vm/val.rs
jacob-hughes: Fixed: 311a7472e5b77a3f7ceb3e9d932834c11ec07bc0

src/lib/vm/val.rs
jacob-hughes: Fixed: 311a7472e5b77a3f7ceb3e9d932834c11ec07bc0

lang_tests/system_global_lookup_string.som
jacob-hughes: Fixed: 2491270453242427a3ed1a9c5f43ee613e94f011

jacob-hughes:
We now explicitly switch out the allocator in yksom's `main.rs` file. The only other proper change we need to make is to ensure that Rust runs our unit tests serially (597780e4d8a27de0808ecef03f0ceff625619519)

ltratt:
Please squash.

jacob-hughes:
Squashed

jacob-hughes:
I'm surprised this merged. I had just noticed that it was a draft PR. I assumed it would need upgrading beforehand.

PR #110, opened 2020-04-27
Remove clones

ltratt:
This PR removes many unnecessary `clone` calls now that `Gc` is `Copy` (which allows us to make `Val` `Copy` too). This is a mix of automatic and manual work.


ltratt:
bors try

PR #111, opened 2020-04-27
Change our sole use of Rc to Gc.

ltratt:
[This needs https://github.com/softdevteam/rboehm/pull/9 to be merged first.]

We originally used Rc here because in abgc Gc<String> was not supported; rboehm
does support this, so we can drop the use of Rc.


jacob-hughes:
I'm afraid I'm not sure what `send` and `reverse_send`s are so can't comment on whether it's an appropriate thing to make garbage-collected. The code LGTM though.

PR #113, opened 2020-05-18
Remove the is_str attribute from String_

ltratt:
SOM has `String` and `Symbol` types where the latter is a subclass of the former. This is a bit awkward to model in Rust. Previously `String_`s had an `is_str` boolean (true for `String`, false for `Symbol`), which meant that we ended up branching on it quite often.

This PR changes that so the creator of a string has to specify if they're creating a SOM `String` (`String_::new_str`) or a `Symbol` (`String_::new_sym`), allowing us to embed the appropriate class in the `String_` struct itself. This doesn't increase the size of the struct (since the `is_str` bool was rounded up to a machine word anyway), and it removes all the subsequent branches.

This does however make bootstrapping the VM a bit messier. The last two commits in this PR try to make that part of the code a little less awkward to read.

### DIFF
--- a/lib/SOM/Symbol.som
+++ b/lib/SOM/Symbol.som
@@ -2,5 +2,5 @@ Symbol = String (
     asString = primitive
     asSymbol = ( ^self )
     "FIXME implement super keyword instead"
-    print    = ( '#' print. (self asString) print )
+    print    = ( '#' print. system printString: self )
 )
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -170,7 +170,7 @@ impl<'a, 'input> Compiler<'a, 'input> {
             return Err(errs);
         }
 
-        let name_val = String_::new(vm, name, false);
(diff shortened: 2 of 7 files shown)


smarr:
Not sure what's messy about the bootstrap, yeah, there's quite a bit that needs setting up.

I only looked briefly, looked sensible to me.

ltratt:
Thanks for taking a look @small! I meant "messy" in the sense of "more code that has to execute in a very precise order, and it's difficult for people later to work out why." In retrospect, perhaps "intricate" or "involved" might have been a better term.

smarr:
Yeah, I agree, it's messy in that sense.

But no need to call me small :rage4: :satisfied:

ltratt:
Mea culpa! That is a hell of a typo (I blame GitHub's autocorrect of usernames). Fixed and hopefully the user who was randomly mentioned isn't too confused :p

smarr:
No worries just had fun exploring github emojis

ptersilie:
Looks good.

bors r+

PR #114, opened 2020-05-19
Raise errors when passing a string instead of a symbol

ltratt:
This PR has a simple goal in mind (`global:s` should raise an error if `s` is anything other than a SOM `Symbol`), but has to go through some hoops to get there. It might be easiest to start with the last commit (https://github.com/softdevteam/yksom/commit/6f14292fc172cd739924a222ef5d11a02d45e810) to see what the goal is.

The challenge is then set out in the commit message https://github.com/softdevteam/yksom/commit/fc3f3f9eebf0fa4c4ba449d98cce2f08c8baf2a0: in essence, at least internally to the VM, there are two different kinds of type errors, and previously we only supported one. So this PR adds a second kind (after renaming things a bit in https://github.com/softdevteam/yksom/commit/c569349966d2e5c709dca5f720d0cfb74b1f9d7e) so that, ultimately, we can raise the error message you see in https://github.com/softdevteam/yksom/commit/6f14292fc172cd739924a222ef5d11a02d45e810. This probably makes it sound more complicated than it is -- sorry!

### DIFF
--- a/lang_tests/system_global_lookup_string.som
+++ b/lang_tests/system_global_lookup_string.som
@@ -1,15 +1,14 @@
 "
-ignore: currently symbols are not properly handled
 VM:
   status: error
   stderr:
-    xxx
+    ...
+    Expected instance of 'Symbol' but got instance of 'String'.
 "
 
 
-system_global_unset_global = (
+system_global_lookup_string = (
     run = (
         (system global: 'ab') println.
     )
 )
-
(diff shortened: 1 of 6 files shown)


ptersilie:
Looks good. Only comment is that we could shorten the last commit message to: ` Add a type error and test for global:put: string (instead of a symbol)`. Though the commit message is also fine as it is. I'll leave it up to you if you want to change it.

ltratt:
Agreed. Should I force push?

ptersilie:
Go ahead.

ltratt:
Force pushed.

PR #115, opened 2020-05-20
super sends

ltratt:
In SOM, "super" is mostly an alias for "self" *except* when you get a message send such as "super x" which means "call the method 'x' in my superclass (or, if not found, recursively in its superclasses)". This allows us to make another part of Symbol.som identical to the SOM standard library.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -470,7 +470,7 @@ impl VM {
                 Instr::Return => {
                     return SendReturn::Val;
                 }
-                Instr::Send(send_idx, cache_idx) => {
+                Instr::Send(send_idx, cache_idx) | Instr::SuperSend(send_idx, cache_idx) => {
                     let (send_rcv, nargs, meth) = {
                         debug_assert!(send_idx < self.sends.len());
                         let nargs = unsafe { self.sends.get_unchecked(send_idx) }.1;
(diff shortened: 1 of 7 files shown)


src/lib/vm/core.rs
```diff
@@ -485,7 +485,15 @@ impl VM {
                                 // The inline cache is empty or out of date, so store a new value in it.
                                 let cls: &Class = stry!(rcv_cls.downcast(self));
                                 let name = unsafe { self.sends.get_unchecked(send_idx) }.0;
-                                let meth = stry!(cls.get_method(self, &*name));
+                                let meth = match instr {
+                                    Instr::Send(_, _) => stry!(cls.get_method(self, &*name)),
+                                    Instr::SuperSend(_, _) => {
+                                        let supercls_val = cls.supercls(self);
```
smarr, line 491: This doesn't look right. `super` sends are lexically bound.

So, you need to know the super class of the method where you are statically in. Not the super of the receiver.

smarr:
Sorry for being late. I am in marking hell. But this is not a proper super semantics.

ltratt:
OK, I must admit I could not for the life of me understand how this was implemented in Java SOM. I'll see if I can fix this.

ltratt:
@small Do you have any way to write a test for this? The problem I have is that SOM only seems to allow 1 class per file, which makes writing single-file tests challenging. Maybe I just have to give up and have multi-file tests...

smarr:
You'll have to have multi-file tests, sorry...
Tests are here: https://github.com/SOM-st/SOM/blob/master/TestSuite/SuperTest.som

It may not actually catch your bug though.

ltratt:
OK, I'll come up with a test that can capture this.

smarr:
The Java implementation does:

1. get the signature as bytecode argument: https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Interpreter.java#L135
2. from the activated method, i.e., the currently executing one (which corresponds necessarily to the lexical one) get the class (holder), and get its superclass https://github.com/SOM-st/som-java/blob/master/src/som/interpreter/Interpreter.java#L139
3. start the lookup in that class

ltratt:
Cool, thanks . I think such a change should be pretty simple.

smarr:
To see it fail, you need a test where you start to low in the hierarchy to make the lookup.

Untested, but something like this perhaps:
```
A = (
  do = ( ^ #a )
)

B = A (
  do = ( ^ #b )
  test = ( ^ super do )
)

C = A (
  do = ( ^ #c )
)

C new test == #a
```

ltratt:
@smarr Yes, that test almost works, it just needs `C = A` changing to `C = B`. A PR to fix this is coming imminently. Thanks for the test case!

PR #116, opened 2020-05-20
Lexical super sends

ltratt:
This PR fixes a problem that @smarr noticed in https://github.com/softdevteam/yksom/pull/115: super sends in yksom were dynamically and not lexically scoped.

In order to do that we first have to allow loading arbitrary user classes. That means adding the directory of the source file to the classpath and allowing references to super classes to load a user class (https://github.com/softdevteam/yksom/commit/bbe8cc2e6350e7f2331d83a91a3ed8ed5fbd4b18). With that done, lexical super sends are relatively easy (https://github.com/softdevteam/yksom/commit/1bbae018c2d524be7ae91ed7d2a2b363b085c172), although we also need to add the ability to have multi-file tests.

[This PR is a bit nicer with https://github.com/softdevteam/lang_tester/pull/58 but it's not strictly necessary.]


lang_tests/lexical_super/test.som
```diff
@@ -0,0 +1,12 @@
+"
+VM:
+  status: success
+  stdout:
+    #a
+"
+
+test = (
+    run = (
+        C new test == #a println.
```
smarr, line 10: Your test is broken. Unary messages have highest priority.
You'll need parentheses:

```suggestion
        (C new test == #a) println
```

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -51,12 +51,9 @@ impl<'a, 'input> Compiler<'a, 'input> {
         let name = lexer.span_str(astcls.name).to_owned();
         let (supercls, supercls_meta) = if name != "Object" {
             let supercls = if let Some(n) = astcls.supername.map(|x| lexer.span_str(x)) {
-                match n {
-                    "Block" => vm.block_cls,
```
smarr, line 55: Nice :)

src/lib/vm/core.rs
```diff
@@ -963,6 +963,24 @@ impl VM {
...
+    /// a) try to load a class `name.som` b) if successful set that as the global `name` c) return
+    /// the class's [Val].
+    pub fn get_global_or_load_class(&mut self, name: &str) -> Result<Val, ()> {
+        if let Some(i) = self.reverse_globals.get(name) {
+            let v = self.globals[*i];
+            if v.valkind() != ValKind::ILLEGAL {
+                return Ok(v);
+            }
+        }
+        if let Ok(p) = self.find_class_path(name) {
```
smarr, line 976: Next PR, support for `Object>>#unknownGlobal:` https://github.com/SOM-st/SOM/blob/master/Smalltalk/Object.som#L81 :)

ltratt: We do support `unknownGlobal`, but I had assumed that it shouldn't be called when looking up a class's superclass, otherwise bootstrapping would become a total mess? Perhaps I'm wrong about that though?

lang_tests/lexical_super/test.som
ltratt: Good point. Slightly different fix, but similar intent in https://github.com/softdevteam/yksom/pull/116/commits/8f11eb83612e3be7b9dc64740a1f926168f6411f.

src/lib/vm/core.rs
smarr: Hm, ok. I think I was just confused by the naming of the method. I did assume this one is used for all `get_global`, but it doesn't seem to be the case.

ltratt: I agree, the name was not well chosen. The next PR will fix this so that the naming is closer to SOM conventions.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #119, opened 2020-05-22
Integer simplifications

ltratt:
This PR makes two simplifications to yksom's internals that I've realised will be useful based on another PR I'm currently working on -- in both cases, the APIs were created very early in yksom's development when we weren't quite sure what the right shape of things should be. The commit messages hopefully make clear the motivation for each change.


ltratt:
bors try

PR #120, opened 2020-05-23
Implement Double.asInteger.

ltratt:
Our implementation is rather inefficient, since we convert immediately to a BigInt, but it is simple and hopefully correct.

### DIFF
--- a/lang_tests/double_asinteger.som
+++ b/lang_tests/double_asinteger.som
@@ -1,0 +1,26 @@
+"
+VM:
+  status: success
+  stdout:
+    0
+    0
+    1
+    0
+    0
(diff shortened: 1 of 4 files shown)


ltratt:
See also https://github.com/SOM-st/SOM/issues/34.

ltratt:
This will need https://github.com/softdevteam/yksom/pull/119 merged first, and then this will need squashing before merging.

ltratt:
@ptersilie If you're OK with this, I'll need to squash before merging.

ptersilie:
Please do.

ltratt:
Force pushed.

PR #122, opened 2020-05-24
Add placeholders for the remaining SOM globals.

ltratt:
This now means we compile the full set of globals (though, obviously, quite a few do not yet have a run-time implementation!).


ptersilie:
Looks good.

PR #123, opened 2020-05-24
Fix incorrect asserts.

ltratt:
These asserts lacked tests that caused them to fire, and hence the fact that they were incorrect had slipped through the net. Oops. Ironically, the asserts were wrong because I overthought them: the whole point of this assert is to check that we only make `ArbInt`s for numbers that can't fit in an `isize`, so each assert is identical.


ptersilie:
Looks good.

PR #124, opened 2020-05-24
Implement Arrays

ltratt:
This implements support for arrays, including array literals.


lang_tests/array_at_negative_idx_err.som
```diff
@@ -0,0 +1,17 @@
+"
+VM:
+  status: error
+  stderr:
+    Traceback...
+    ...
+    Can't represent as unsigned machine integer.
```
smarr, line 7: Error doesn't seem to be very helpful.

lang_tests/array_at_idx_err.som
```diff
@@ -0,0 +1,17 @@
...
+    ...
+    Index 1 exceeds array length 0.
+"
+
+array_at_idx_err = (
+    run = (
+        | x |
+
+        x := Array new.
+        x at: 1.
```
smarr, line 15: What do you get when accessing index 0?

lang_tests/class_load/test.som
```diff
@@ -2,11 +2,11 @@
...
   stdout:
-    Array
+    other
 "
 
-class_load = (
+test = (
     run = (
-        Array println.
+        other new f println.
```
smarr, line 10: IMHO, this code is hard to read, because I have less info about what `other` is than what I would have if Smalltalk conventions were followed. `Other` tells me it's a class, very likely. Especially when it is in this specific expression, because it has to be some kind of read.

smarr: And of course, you have stdlib classes capitalized and your classes not, which is a strange inconsistency.

ltratt:
Embarrassingly I forgot to add the test file for array literals: fixed in https://github.com/softdevteam/yksom/pull/124/commits/8ef1f28d0399279d9d8bbf015df8f3ea2cda5dc3.

lang_tests/array_at_idx_err.som
ltratt: Mea culpa! Yes, this was very broken. Fixed, tests added, (and I also implemented indexing exceptions for `at:put:`) in https://github.com/softdevteam/yksom/pull/124/commits/33a37e9a3631f5aef041f425ea55d4c4ef96421e.

lang_tests/array_at_negative_idx_err.som
ltratt: It's not ideal, but doing something better requires a bit more work than I want to do in this PR.

ptersilie:
Looks good. Please squash.

ltratt:
Oops, I forgot to push a very-very-minor commit https://github.com/softdevteam/yksom/pull/124/commits/82014741b2abe16a25d3067f66d0774a63743db5, but I don't want to sneak it into the squash. If you're OK with it, I can then squash. Sorry!

ptersilie:
No problem. You can squash now.

ltratt:
Squashed.

PR #126, opened 2020-05-28
Adjust to the new natrob API that returns Gc from downcast.

ltratt:
This reflects the change in https://github.com/softdevteam/natrob/pull/14#issuecomment-635622030.

### DIFF
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -179,7 +179,7 @@ impl<'a, 'input> Compiler<'a, 'input> {
             methods,
         );
         let cls_val = Val::from_obj(vm, cls);
-        let cls: &Class = cls_val.downcast(vm).unwrap();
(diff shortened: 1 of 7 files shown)


jacob-hughes:
All LGTM. As a general observation with this stuff, It looks like you will (or are just about to) run into the need for arbitrary self types. I am pretty sure I did this when doing the API for gcmalloc (but can't remember now). Has that been working ok?

ltratt:
I don't *think* I'll need arbitrary self types here (though I would love them to work fully in some other contexts!).

jacob-hughes:
Cool. I'm interested to see how well it works.

ltratt:
> Cool. I'm interested to see how well it works.

So am I!

ltratt:
A reminder to self: test everything fully before making a PR. The problem that caused the buildbot failure is fixed in https://github.com/softdevteam/natrob/pull/15. Once that's merged, this PR should build.

PR #130, opened 2020-06-04
Invalidate inline caches on method mutation

ltratt:
This fixes the problem identified in https://github.com/SOM-st/SOM/issues/35 and which, AFAIK, affects all SOM implementations: mutating a method in a class causes incorrect behaviour if an inline cache exists for that call (e.g. yksom (before this PR) and Java SOM could "ignore" the change if a cache had been created; and C SOM apparently crashes!). This PR fixes the problem for yksom.

It does so in a brute force way: when a method in any class is changed, every inline cache in the VM is reset. This makes correctness easy to reason about, though clearly it is less efficient than more nuanced approaches.

The way we do this is sort-of interesting from a Rust perspective. There are two obvious ways to model this:

* Have every SOM `Array` have a `bool` that says "if I'm mutated, flush all inline caches".
* Subclass the `Array` struct (in Rust) so that only instances of the `MethodArray` subclass flush inline caches on mutation.

The former is ugly, the latter seems impossible: Rust doesn't have subclasses, and traits only partially model them. In this case, the `bool` would probably be most efficient, but it's just really ugly. So I tried to make the latter case work, even though it doesn't appear to be possible. In particular, and perhaps surprisingly (though there are good reasons for this), a trait object `T1 + T2` cannot be coerced into a trait object of either `T1` or `T2`. This PR uses a simple way to solve this problem:

1. We define a trait `Array`
2. We define two structs `NormalArray` (which is the old `Array` struct renamed) and `MethodsArray` (which is only used for the arrays of methods in `Class`es).
3. We add a `to_array` function to `Obj` which returns (simplified) `&dyn Array`. Given a `&dyn Obj` we can then turn it into a `&dyn Array` by calling `to_array` *if and only if* an implementation of `to_array` is provided. Only `NormalArray` and `MethodsArray` implement this.

Using this, we can get from a `Val` to a `&dyn Array` by calling (simplified) `val.tobj().to_array()`. Yes, it means we have two dynamically dispatched function calls, but it means that we have a relatively nice way of separating out the behaviour of normal arrays from method arrays.

Most of the PR is thus shuffling code around (and duplicating some code): the real meat is in https://github.com/softdevteam/yksom/commit/0889b12e8025dc56cb8aaaf0d8312a6af3e980f6.

### DIFF
--- a/lang_tests/mutate_methods.som
+++ b/lang_tests/mutate_methods.som
@@ -1,0 +1,24 @@
+"
+VM:
+  stdout:
+    #b
+    #c
+"
+
+mutate_methods = (
+    run = (
+        | g_idx h_idx ms |
+        self f println.
+        ms := mutate_methods methods.
+        ms doIndexes: [:i |
+            (((ms at: i) signature) == #g) ifTrue: [ g_idx := i ].
+            (((ms at: i) signature) == #h) ifTrue: [ h_idx := i ].
+        ].
+        mutate_methods methods at: g_idx put: (ms at: h_idx).
+        self f println.
+    )
+
+    f = ( ^self g )
+    g = ( ^#b. )
+    h = ( ^#c. )
+)
--- a/src/lib/vm/objects/array.rs
+++ b/src/lib/vm/objects/array.rs
@@ -9,6 +9,12 @@ use crate::vm::{
     val::{NotUnboxable, Val},
 };
 
+pub trait Array {
+    fn at(&self, vm: &VM, idx: usize) -> Result<Val, Box<VMError>>;
+    unsafe fn unchecked_at(&self, idx: usize) -> Val;
+    fn at_put(&self, vm: &VM, idx: usize, val: Val) -> Result<(), Box<VMError>>;
+}
+
 #[derive(Debug)]
 pub struct NormalArray {
     store: UnsafeCell<Vec<Val>>,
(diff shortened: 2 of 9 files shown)


lang_tests/mutate_methods.som
```diff
@@ -0,0 +1,24 @@
+"
+VM:
+  stdout:
+    #b
+    #c
+"
+
+mutate_methods = (
+    run = (
+        | g_idx h_idx ms |
```
smarr, line 10: I will not stop to point out that changing coding style amounts to changing the language...

Smalltalks use camelCase for variables, method names, ...
They use PascalCase for class names.

Inconsistency has a non-zero cost.

src/lib/vm/objects/array.rs
```diff
@@ -88,11 +181,12 @@ impl Array {
 
     /// Set the item at index `idx` (using SOM indexing starting at 1) to `val` or return an error
     /// if the index is invalid.
-    pub fn at_put(&self, vm: &VM, mut idx: usize, val: Val) -> Result<(), Box<VMError>> {
+    fn at_put(&self, vm: &mut VM, mut idx: usize, val: Val) -> Result<(), Box<VMError>> {
         let store = unsafe { &mut *self.store.get() };
         if idx > 0 && idx <= store.len() {
             idx -= 1;
             *unsafe { store.get_unchecked_mut(idx) } = val;
+            vm.flush_inline_caches();
```
smarr, line 189: A cheap trick could be `if (old != new) flushCache`.

ltratt: Agreed, but this is such a crude mechanism (and `old == new` an unlikely outcome) that I'm not sure it's worth the complexity. When we eventually JIT this VM I hope to use a much more fine-grained mechanism, but it feels like overkill in a plain interpreters!

lang_tests/mutate_methods.som
ltratt: If someone makes a PR to change the whole lot in one go, I'll happily change style to match other SOM's conventions henceforth. Until then, I'm just going to carry on being consistent internally within yksom.

src/lib/vm/objects/array.rs
```diff
@@ -29,6 +29,10 @@ impl Obj for NormalArray {
         vm.array_cls
     }
 
+    fn to_array(&self) -> Result<&dyn Array, Box<VMError>> {
```
ptersilie, line 32: Why does this return a `Result`. I don't see how this can cause an error here.

ptersilie:
There's one (possibly two) typos in the commit messages:
- `to do solve`: 1a86c05
- `#b #b`: 0889b12

Everything else looks fine. So feel free to force-push.

ltratt:
> to do solve: 1a86c05

Fixed.

> #b #b: 0889b12

That one is correct (it's showing a bug in Java SOM).

I've forced pushed an update.

src/lib/vm/objects/array.rs
ltratt: It has to implement with this return type because it's defined in the trait.

PR #132, opened 2020-06-04
Implement string length and substring

ltratt:


src/lib/vm/objects/string_.rs
```diff
@@ -128,6 +128,25 @@ impl String_ {
         Ok(String_::new_str(vm, new))
     }
 
+    pub fn substring(&self, vm: &mut VM, start: usize, end: usize) -> Result<Val, Box<VMError>> {
+        if start == 0 || end == 0 {
```
ptersilie, line 132: What about `start < 0`?

ltratt: They're `usize` so they can't be `< 0`.

ptersilie: Oh I missed that.

PR #136, opened 2020-06-05
Call escapedBlock:

ltratt:
This PR calls `escapedBlock:` when a block escapes (i.e. when it tries to return a value even though the closure it was called within has already been returned from).

### DIFF
--- a/lang_tests/escaped1.som
+++ b/lang_tests/escaped1.som
@@ -1,9 +1,8 @@
 "
 VM:
   status: error
-  stderr:
-    ...Return from escaped block...
-    ...
+  stdout:
+    ERROR: Block has escaped and cannot be executed
 "
 
 escaped1 = (
--- a/lang_tests/escaped2.som
+++ b/lang_tests/escaped2.som
@@ -1,0 +1,18 @@
+"
+VM:
+  status: success
+  stdout:
+    instance of Block1
+"
+
+escaped2 = (
+    f = ( ^[ ^ 'a' ] )
+
+    run = (
+        self f value
+    )
+
+    escapedBlock: block = (
+        block println.
+    )
+)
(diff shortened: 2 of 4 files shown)


smarr:
The PR description uses the word closure in a way that may indicate a bug.

What does the following give:

```Smalltalk
BlockInBlock = (

  method = (
    | block |
    block := [
      | var |
      var := 22.
      [ ^ var ]
    ] value.
    
    block value
  )

  run = (
    'I expect 22, and get: ' print.
    self method println.
  )

)
```

ltratt:
> The PR description uses the word closure in a way that may indicate a bug.

That wasn't the intention (at least not if I understand you correctly; though I can't imagine a case where an escaped block isn't a bug).

Your example prints `I expect 22, and get: 22` in this PR.

ltratt:
> That wasn't the intention (at least not if I understand you correctly; though I can't imagine a case where an escaped block isn't a bug).

Ah, I see what you mean now! You thought (not unreasonably) that I hadn't described things properly. Point taken: it's actually a bit difficult to find the right terminology for this in SOM. Internally yksom in essence calls an instance of a method a "closure", which might explain things.

smarr:
Yepp, was just going to clarify.

You got two closures here, the method, and the outer block.
And, indeed, we already returned from one closure in the example, but not from the method.
So, here, I'd always use the term method to avoid confusion.

ltratt:
The challenge with terminology is that I can create a method with an inner block, on one invocation of which the block escapes and another it doesn't. They could even be on the stack. Finding the right terms is difficult! Anyway, I think we're on the same page here which is good.

smarr:
> I can't imagine a case where an escaped block isn't a bug

As "escaped", I'd refer to any block that is returned from the methods it was defined in (i.e. the method is not on the stack anymore).

Such blocks are commonly used as callbacks, and are unproblematic as long as they do not use explicit returns. They are extremely common in JavaScript, and common enough in Smalltalk. Though, yeah, a good fraction of all blocks in probably only ever used in the dynamic scope of their method.

ltratt:
You're right, I was sloppy with terminology again! I meant "escaped" as in "tried to do a return after escaping" (in the sense of `escapedBlock:`).

PR #139, opened 2020-06-07
Array literals don't have to have any items in them.

ltratt:
The grammar previously required, incorrectly, at least one item in array literals.


ltratt:
@ptersilie I think we killed buildbot while it was running, so you hopefully just need to re-rplus this.

PR #140, opened 2020-06-08
Implement the 'fullGC' primitive.

ltratt:
We do not actually do a GC, since a) we have no way of forcing one b) asking a VM to GC should be a hint rather than a request so not doing anything is a valid thing to do anyway.

### DIFF
--- a/lang_tests/fullgc.som
+++ b/lang_tests/fullgc.som
@@ -1,0 +1,11 @@
+"
+VM:
+  status: success
+  stdout: false
+"
+
+fullgc = (
+    run = (
+        system fullGC println.
+    )
+)
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -719,7 +719,10 @@ impl VM {
             }
             Primitive::Fields => todo!(),
             Primitive::FromString => todo!(),
-            Primitive::FullGC => todo!(),
+            Primitive::FullGC => {
+                self.stack.push(self.false_);
+                SendReturn::Val
+            }
             Primitive::Global => {
                 let name_val = self.stack.pop();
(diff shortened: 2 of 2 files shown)


smarr:
It's an `expect:` test, which is:

https://github.com/SOM-st/SOM/blob/master/TestSuite/TestCase.som#L28-L30

```
    expect: aBoolean = (
        aBoolean ifFalse: [
            self signalInefficacy: 'Expectation unsatisfied' ] )
```

Not being a native speaker, I chose the term inefficacy for test failures for things that are desirable to have. 

Using the terminology of https://www.ietf.org/rfc/rfc2119.txt
an `assert:` is a MUST, and a `expect:` is a SHOULD.

```
1. MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
   definition is an absolute requirement of the specification.

3. SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.
```

ltratt:
Ah, I see. I must admit, I did wonder what "inefficacy" meant (literally: I didn't know the word at all!). I'm all for short words myself: you could even say "X 'must pass' tests failed" and "Y 'optional' tests failed" or something like that? [In my experience, people tend not to differentiate "must" and "should" unless they're explicitly told to e.g. in a standards document.]

ltratt:
@ptersilie If you're happy with that new commit, I should squash before any potential merge.

ptersilie:
Looks good. Please squash.

ltratt:
Squashed (and updated PR description).

PR #145, opened 2020-06-08
Partiallly implement the 'rem:' operator.

ltratt:
This calculates remainders. Fully implementing this for all the possible types is surprisingly tricky because BigInt doesn't provide `checked_rem`. Since Java SOM also can't deal with any of the tricky cases (a big int / non-big int errors with "TODO"), I think the very basic support here is just about acceptable.

### DIFF
--- a/lang_tests/remainder.som
+++ b/lang_tests/remainder.som
@@ -1,0 +1,14 @@
+"
+VM:
+  status: success
+  stdout:
+    1
+    2
+"
+
+remainder = (
+    run = (
+        (10 rem: 3) println.
+        (20 rem: 3) println.
+    )
+)
--- a/lang_tests/remainder_zero.som
+++ b/lang_tests/remainder_zero.som
@@ -1,0 +1,14 @@
+"
+VM:
+  status: error
+  stderr:
+    Traceback...
+    ...
+    Division by zero or overflow.
+"
+
+remainder_zero = (
+    run = (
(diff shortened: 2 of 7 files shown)


vext01:
This seems reasonable to me.

Hows about some tests?

ltratt:
I'm an idiot: I forgot to add the test file. Pushed.

lang_tests/remainder.som
```diff
@@ -0,0 +1,15 @@
...
+  stdout:
+    1
+    2
+"
+
+remainder = (
+    run = (
+        (10 rem: 3) println.
+        (20 rem: 3) println.
+        ((1 << 100) rem: 3) println.
```
vext01, line 13: Are you able to test for failures? If so, you've identified 2 that would be worth in including:
 - divide by zero,
 - overflow.

vext01: Is the third case a failure actually? There are only two expected outcomes.

ltratt: I believe, ironically, that the overflow case can't happen with oversize rem: oversize. But you're right about divide by zero. The third test is also bogus at the moment. All fixed in https://github.com/softdevteam/yksom/pull/145/commits/0ddde6ca5544228576575f41494c248e7547353d.

vext01:
Cool, please squash.

ltratt:
Squashed.

PR #146, opened 2020-06-09
Implement the 'as32Bit[Un]signedValue' primitives.

ltratt:
Java SOM crashes on big integers, so I have defined what I believe to be appropriate semantics for them which appears to be not to perform sign extension in the conversion. For example:

```
  ((1 << 60) - 1) as32BitUnsignedValue
```

in Java SOM returns 4294967295, which is bigger than a signed 32 bit can represent. This commit preserves that behaviour.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -636,8 +636,34 @@ impl VM {
                 self.stack.push(v);
                 SendReturn::Val
             }
-            Primitive::As32BitSignedValue => todo!(),
-            Primitive::As32BitUnsignedValue => todo!(),
+            Primitive::As32BitSignedValue => {
+                let i = if let Some(i) = rcv.as_isize(self) {
+                    i as i32 as isize
+                } else {
+                    rcv.downcast::<ArbInt>(self)
+                        .unwrap()
+                        .bigint()
+                        .to_u32_digits()
+                        .1[0] as i32 as isize
+                };
+                let v = Val::from_isize(self, i as isize);
+                self.stack.push(v);
+                SendReturn::Val
+            }
+            Primitive::As32BitUnsignedValue => {
+                let i = if let Some(i) = rcv.as_isize(self) {
+                    i as u32
+                } else {
+                    rcv.downcast::<ArbInt>(self)
+                        .unwrap()
+                        .bigint()
+                        .to_u32_digits()
+                        .1[0] as u32
+                };
+                let v = Val::from_isize(self, i as isize);
+                self.stack.push(v);
+                SendReturn::Val
+            }
             Primitive::At => {
                 let rcv_tobj = stry!(rcv.tobj(self));
                 let arr = stry!(rcv_tobj.to_array());
--- a/lang_tests/to32bits.som
+++ b/lang_tests/to32bits.som
@@ -1,0 +1,43 @@
+"
+VM:
+  status: success
+  stdout:
+    0
+    1
+    4294967295
+    4294967295
+    0
+    4294967295
+    0
+    4294967295
+    0
+    1
+    -1
+    -1
+    0
+    -1
+    0
+    -1
+"
+
+to32bits = (
+    run = (
+        0 as32BitUnsignedValue println.
+        1 as32BitUnsignedValue println.
+        -1 as32BitUnsignedValue println.
+        4294967295 as32BitUnsignedValue println.
+        (1 << 60) as32BitUnsignedValue println.
+        ((1 << 60) - 1) as32BitUnsignedValue println.
+        (1 << 200) as32BitUnsignedValue println.
+        ((1 << 200) - 1) as32BitUnsignedValue println.
+
+        0 as32BitSignedValue println.
+        1 as32BitSignedValue println.
+        -1 as32BitSignedValue println.
+        4294967295 as32BitSignedValue println.
+        (1 << 60) as32BitSignedValue println.
+        ((1 << 60) - 1) as32BitSignedValue println.
+        (1 << 200) as32BitSignedValue println.
+        ((1 << 200) - 1) as32BitSignedValue println.
+    )
+)


vext01:
Before I look at the code, let me get my head around what's going on here with some (probably dumb) questions.

>   ((1 << 60) - 1) as32BitUnsignedValue

What type are integer constants in SOM?

ltratt:
You're asking me to justify SOM's language design, and I can't justify these particular features: they seem wrong to me, both in principle and in their implementation. I assume they result from the first SOM being developed on a 32-bit machine. All I have tried to do is to maintain Java SOM's behaviour.

vext01:
> You're asking me to justify SOM's language design

I'm not asking you to justify it, just tell me how they've done it.

> ((1 << 60) - 1) as32BitUnsignedValue

In this expression, is `((1 << 60) - 1)` a signed 32-bit integer?

ltratt:
> I'm not asking you to justify it, just tell me how they've done it.

No, you really are asking me to justify it because "how they've done it" means understanding the language design and I don't understand these particular features :) I don't know how other SOMs work internally, other than Java SOM has the same observable behaviour as this PR, presumably because its ints are also 64-bits. I can only repeat myself and say that the tests run the same on Java SOM and yksom (other than the "big" ints `1 << 200` where Java SOM throws a backtrace because it misses those cases).

vext01:
You mentioned that SOM doesn't sign extend in the PR description.

Unless I'm wrong, to observe that sign extension doesn't happen, you'd have to cast a small negative number to a wider-sized negative number. Do we need to add a test for that? Or do I misunderstand?

ltratt:
> Unless I'm wrong, to observe that sign extension doesn't happen, you'd have to cast a small negative number to a wider-sized negative number.

I'm not sure what you mean. Do you want to give me a concrete example?

vext01:
Sure.

Usually sign extension refers to preserving a negative value during a cast to a wider signed type.

```
fn main() {
    let x = -4i8;
    
    println!("{:b}", x);
    println!("{:b}", x as i16);
}
```

Gives:
```
11111100
1111111111111100
```

And the sign-extend was the left padding with 1s to ensure the value remains -4.

It's not clear to me which test shows that this doesn't happen in SOM.

ltratt:
I wonder if you might have misread the PR description?

> I have defined what I believe to be appropriate semantics for them which appears to be **not** to perform sign extension in the conversion

I have highlighted what I think is the key word!

ltratt:
Hang on, some of the tests use the wrong function.

ltratt:
I am an idiot: I copied and pasted tests and forgot to change the function that was called. That led me to completely misunderstand what Java SOM was doing and unfairly malign SOM -- sorry @smarr! @vext01 your prodding was very helpful in making me check this again. https://github.com/softdevteam/yksom/pull/146/commits/02c5f94c7b0255c3760d28aa9ead852e5efbb280 fixes things.

vext01:
OK, that makes a little more sense now.

If `(1 << 60) as32BitUnsignedValue` is 0 (presumably because the `1` fell of the left-hand side of the 32-bit integer), then `((1 << 60) - 1)` underflows the value to `-1`, and then the cast `as32BitSignedValue` is then a NOP (sign extend is irrelevant in this case I think).

I'm happy to merge this if this matches other implementations.

Do you want to squash.

ltratt:
I agree with your updated explanation and this time I've checked more carefully that the tests I'm executing do what I expect them to!

Squashed.

smarr:
> OK, that makes a little more sense now.
> 
> If `(1 << 60) as32BitUnsignedValue` is 0 (presumably because the `1` fell of the left-hand side of the 32-bit integer), then `((1 << 60) - 1)` underflows the value to the largest signed integer (all ones), and then the cast `as32BitSignedValue` makes -1 (sign extend is irrelevant in this case I think).

So, starting reading from the top, I was very confused by some of the explanations. But here, you seem to agree on things, so, I'll start here.

SOM has arbitrary precision integers (in theory, in practice there are incomplete implementations and bugs).

The literal `1` is an arbitrary precision integer, as is `60`. Usually, these would be represented as 32bit or 64bit integers in implementations.

`1 << 60` should give `1152921504606846976` if I haven't done something silly on the calculator. That's still an arbitrary precision integer, which fits into 64bits.

Subtracting `1` should give `1152921504606846975`. Still a positive number.
What might be relevant here is that the bit pattern is `00001111...`, otherwise, nothing special, I can see. There's no underflow as in "wrap around" going on here.

And then there's confusion on top of confusion. Do you care about `as32BitUnsignedValue` or `as32BitSignedValue`?

```
---> ((1 << 60) - 1) as32BitUnsignedValue
it = 4294967295
---> ((1 << 60) - 1) as32BitSignedValue
it = -1
```

The idea here is: take the bit pattern of the current value, look at the least significant 32 bit, and interpreter it as unsigned or signed value.

The documentation says:

```
    as32BitSignedValue   = primitive " returns an int, with the value that a signed 32-bit integer would have"
    as32BitUnsignedValue = primitive " returns an int, with the value that a unsigned 32-bit integer would have"
```

I guess that phrasing can be clarified. Though, not sure how and where the confusion is, because there seem to be various things that don't seem to be clear.

Would be good to know what the assumptions are that underly the current interpretation.

src/lib/vm/core.rs
```diff
@@ -636,8 +636,34 @@ impl VM {
...
+                let i = if let Some(i) = rcv.as_isize(self) {
+                    i as i32 as isize
+                } else {
+                    rcv.downcast::<ArbInt>(self)
+                        .unwrap()
+                        .bigint()
+                        .to_u32_digits()
+                        .1[0] as i32 as isize
+                };
+                let v = Val::from_isize(self, i as isize);
```
smarr, line 649: Small quibble, the `as isize` appears here 3 times, while below only 1 time. So, I'd have thought having it once here would also be fine.

vext01:
This is why I was asking about the type of integer constants to start out with.

So if they are AP, then Stefan's explanation checks out. The truncation caused by the cast then leads to a result that happens to be interpreted as -1.

Demonstrated here in ipython:
```
In [4]: (1<<60) - 1
Out[4]: 1152921504606846975

In [5]: bin(_)
Out[5]: '0b111111111111111111111111111111111111111111111111111111111111'

In [7]: 1152921504606846975 & 0xffffffff
Out[7]: 4294967295

In [8]: bin(_)
Out[8]: '0b11111111111111111111111111111111'
```

And all `1`s is -1 when interpreted signed.

I must emphasize that I know nothing about SOM -- I only draw from experience from other languages I've worked with.

smarr:
> This is why I was asking about the type of integer constants to start out with.

The type is `Integer`, as in arbitrary precision integer.
What you chose to use to represent it is up to you.
SOM doesn't have `int8_t`, `size_t`, `uint64_t` or anything similar.

> So if they are bigint

They are not bigint, they are `Integer`. There's no bigint/smallint divide.

And since you bring up Python, Python seems to do it very similar in that representation is decoupled from semantics, having integers with unlimited precision (as per https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex)

vext01:
To be clear, when I say "bignum", I mean arbitrary precision.

PR #147, opened 2020-06-09
Detect overlapping field names

ltratt:
There are two cases to be dealt with:

1. A field name has already been defined in a superclass (https://github.com/softdevteam/yksom/commit/5e00d91257a8865f90f0836e65bbcddb237078a5).
2. A class defines the same field more than once (https://github.com/softdevteam/yksom/pull/147/commits/93706090edfdcd26f2a0e82032500bd19a0f3686).

The first of those was pointed out by @hirevo in https://github.com/SOM-st/SOM/issues/42; the second is a relatively simple variation on the same theme.

### DIFF
--- a/lang_tests/instance_fields_overlap/test.som
+++ b/lang_tests/instance_fields_overlap/test.som
@@ -1,0 +1,13 @@
+"
+VM:
+  status: error
+  stderr:
+    ...test.som', line 10, column 23:
+      test = test_super ( | a |
+    Field 'a' is already defined in a superclass.
+"
+
+test = test_super ( | a |
+  run = (
+  )
+)
--- a/lang_tests/instance_fields_overlap2.som
+++ b/lang_tests/instance_fields_overlap2.som
@@ -1,0 +1,12 @@
+"
+VM:
+  status: error
+  stderr:
+    ...instance_fields_overlap2.som', line 11, column 9:
+      | a a |
+    Field 'a' has already been defined in this class.
+"
+
+instance_field_overlap2 = (
+    | a a |
+)
--- a/lang_tests/instance_fields_overlap/test_super.som
+++ b/lang_tests/instance_fields_overlap/test_super.som
@@ -1,0 +1,3 @@
+test_super = (
+    | a |
+)
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -153,6 +153,15 @@ impl<'a, 'input> Compiler<'a, 'input> {
                     .iter()
                     .map(|(k, v)| (k.to_owned(), *v)),
             );
+            for var in ast_inst_vars {
+                let n = lexer.span_str(*var);
+                if inst_vars.contains_key(n) {
+                    return Err(vec![(
+                        *var,
+                        format!("Field '{}' is already defined in a superclass.", n),
+                    )]);
+                }
+            }
             inst_vars
         } else {
             HashMap::with_capacity(ast_inst_vars.len())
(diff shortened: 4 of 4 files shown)


lang_tests/instance_fields_overlap/test.som
```diff
@@ -0,0 +1,13 @@
+"
+VM:
+  status: error
+  stderr:
+    ...test.som', line 10, column 23:
+      test = test_super ( | a |
+    Field 'a' is already defined in a superclass.
```
ltratt, line 7: I'm not sure if this should be "Field" or (as below) "Field name". Any thoughts?

smarr: I as non-native speak would say: 'Field a is already defined' or 'A field with the name a is already defined'.
"Field name 'a' has already been defined in this class." feels wrong.
Perhaps 'A field named 'a' ...' as another option.

lang_tests/instance_fields_overlap2.som
```diff
@@ -0,0 +1,12 @@
+"
+VM:
+  status: error
+  stderr:
+    ...instance_fields_overlap2.som', line 11, column 9:
+      | a a |
+    Field name 'a' has already been defined in this class.
```
smarr, line 7: One thing I am surprised by is that you don't format the error message in a "standard way" with path:line:column, which is supported by a large number of tools for navigation to the error.

lang_tests/instance_fields_overlap/test.som
ltratt: Yes, I think on reflection that "Field" is better than "Field name" (due to the `fields` method being a precedent).

lang_tests/instance_fields_overlap2.som
ltratt: I didn't know that's a standard to be honest! Changing the formatting is pretty trivial if anyone wants to raise a PR for it (it's not something I feel very strongly about either way), with the only proviso that I think the console output should still look reasonable for humans.

lang_tests/instance_fields_overlap/test.som
ltratt: Fixed in https://github.com/softdevteam/yksom/pull/147/commits/f57c74b58040796e3a840ea9facd6a552565d775.

vext01: Or even better "A field by the name 'a' is already defined...".

vext01:
A handful of comments.

lang_tests/instance_fields_overlap/test.som
smarr: The most useful improvement would IMHO be to name the superclass and the source location where the conflicting field is defined. That would be very developer-friendly :)

ltratt: I don't think error messages need to use full grammar (so to speak): it can often end up being pointless extra verbiage to plough through especially when you tend to see the same messages over and over again. I think I'd prefer to keep it as in https://github.com/softdevteam/yksom/commit/f57c74b58040796e3a840ea9facd6a552565d775.

ltratt:
I tend to agree with @small that this PR isn't about duplicate superclasses (SOM doesn't have the diamond problem).

vext01:
Agreed.

I think I'm happy with this. Shall I merge?

vext01:
Ah, please squash first.

ltratt:
I need to squash if that's OK?

ltratt:
Squashed.

vext01:
I realise that some of my comments from earlier didn't make a great deal of sense. Maybe another coffee before my next PR review! My bad.

ltratt:
It's not your fault -- I don't expect you to know SOM inside out (I'm only starting to slowly get to grips with it myself!).

PR #148, opened 2020-06-12
Implement the 'Shr' primitive.

ltratt:
Java SOM seems to define "shift right" as "treat the int as unsigned, shift right, then treat as signed". This is fairly easy for us in the case of word-sized integers, but awkward to implement for negative big ints -- I've left that for the future (if anyone should ever encounter such a case, which might be quite some time away!).

### DIFF
--- a/lang_tests/shift_right.som
+++ b/lang_tests/shift_right.som
@@ -1,0 +1,25 @@
+"
+VM:
+  status: success
+  stdout:
+    0
+    512
+    127
+    -1
+    9223372036854775807
+    9223372036854775296
+    1
+"
+
+shift_right = (
+    run = (
+        (1 >>> 1) println.
+        (1024 >>> 1) println.
+        (1023 >>> 3) println.
+        (-1 >>> 0) println.
+        (-1 >>> 1) println.
+        (-1024 >>> 1) println.
+
+        ((1 << 200) >>> 200) println.
+    )
+)
--- a/lang_tests/shift_right_too_big.som
+++ b/lang_tests/shift_right_too_big.som
@@ -1,0 +1,14 @@
+"
+VM:
+  status: error
+  stderr:
+    Traceback...
+    ...
+    Shift too big.
+"
+
+shift_right = (
+    run = (
+        1 >>> (1 << 100).
+    )
+)
--- a/lang_tests/shift_right_type_err.som
+++ b/lang_tests/shift_right_type_err.som
@@ -1,0 +1,14 @@
+"
+VM:
+  status: error
+  stderr:
+    Traceback...
+    ...
+    Expected a numeric type but got type 'String_'.
+"
+
+shift_right = (
+    run = (
+        (1 >>> 'a') println.
+    )
+)
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -973,7 +973,12 @@ impl VM {
                 self.stack.push(v);
                 SendReturn::Val
             }
-            Primitive::Shr => todo!(),
+            Primitive::Shr => {
+                let v = self.stack.pop();
+                let v = stry!(rcv.shr(self, v));
+                self.stack.push(v);
+                SendReturn::Val
+            }
             Primitive::Signature => {
                 let meth = rcv.downcast::<Method>(self).unwrap();
                 self.stack.push(meth.sig(self));
--- a/src/lib/vm/objects/integers.rs
+++ b/src/lib/vm/objects/integers.rs
@@ -193,6 +193,27 @@ impl Obj for ArbInt {
         }
     }
 
+    fn shr(&self, vm: &mut VM, other: Val) -> Result<Val, Box<VMError>> {
+        if let Some(rhs) = other.as_isize(vm) {
+            if rhs < 0 {
+                Err(VMError::new(vm, VMErrorKind::NegativeShift))
+            } else {
+                let rhs_i =
+                    usize::try_from(rhs).map_err(|_| VMError::new(vm, VMErrorKind::ShiftTooBig))?;
+                if self.val.is_positive() {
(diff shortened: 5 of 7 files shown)


vext01:
So we have shift operators with two or three chevrons?

Is there a language reference somewhere for SOM?

ltratt:
There is no language reference for SOM I'm afraid. `<<` is shift left and `>>>` shift right AFAICS. [I suspect the different number of arrows is to differentiate sign extension or similar, but I don't know.]

vext01:
OK.

> "treat the int as unsigned, shift right, then treat as signed"

Does that mean that an unsigned integer whose bit-pattern is all ones, when shifted right by 0 bits, is -1?

lang_tests/shift_right.som
```diff
@@ -0,0 +1,23 @@
...
+"
+
+shift_right = (
+    run = (
+        (1 >>> 1) println.
+        (1024 >>> 1) println.
+        (1023 >>> 3) println.
+        (-1 >>> 1) println.
+        (-1024 >>> 1) println.
+
```
vext01, line 20: extra blank line?

ltratt:
> Does that mean that an unsigned integer whose bit-pattern is all ones, when shifted right by 0 bits, is -1?

That entirely depends on the representation. With twos complement, yes. See https://github.com/softdevteam/yksom/pull/148/commits/09964ce215f4f410e134e1db5d05871d2e44a078.

lang_tests/shift_right.som
ltratt: It's to differentiate machine word sized tests from big int tests (this formatting is used extensively throughout the lang_tests).

vext01: Fair enough.

vext01:
Did you forget to add the expected outcome for the new test?

ltratt:
Mea culpa! Fixed in https://github.com/softdevteam/yksom/pull/148/commits/5a404f2dd31aa3b2dcc4901ecff684c96ea3229a.

vext01:
That's a good test to have, but I was suggesting starting with a positive integer.

For example `4294967295` the largest 32-bit integer, and all ones. If you shift that by 0, do you get -1? In your description, you said the result is always interpreted signed.

It'd be odd...

smarr:
SOM only has signed integers. As discussed earlier, SOM does not have any notion of integer types, but `Integer`. This comes in here of course again, too.

The supported shift operations are exactly what was needed for a benchmark.
Similar to the `as32Bit*` operations.

No existing code will use shift operations on values that don't fit into 64 bit, so, SOM-java never implemented support for it either.

The number of angle brackets is consistent with other languages. Though, as with all these things (worst of all %/rem), people have different opinions...

The commits introducing them are not exactly full of design reasoning:
 - https://github.com/SOM-st/SOM/commit/c9d1b7fcdbba805665073bb40ddf5404ed413e0a
 - https://github.com/SOM-st/SOM/commit/a6b7549639945e0dbe74406af5ef7f2336c23249

But their context tells us that unsigned right shift was introduced for the GraphSearch benchmark, and the signed left shift was introduced for the Mandelbrot benchmark a year earlier.

ltratt:
> For example 4294967295 the largest unsigned 32-bit integer, and all ones. If you shift that by 0, do you get -1?

No, that's impossible for a non-rotating shift right.

smarr:
SOM-java:

```
---> 4294967295 >>> 0
it = 4294967295
---> 4294967295 >>> 1
it = 2147483647
---> 4294967295 << 0
it = 4294967295
---> 4294967295 << 1
it = 8589934590
```

vext01:
> SOM only has signed integers.

OK, so that means that my question is invalid, as any integer with an all one bit pattern is by definition `-1`.

> The number of angle brackets is consistent with other languages.

Honestly I've never seen triple chevrons for shift (that I can remember), but I've only really dealt with mainstream languages.

So be it. We do as SOM does :)

ltratt:
OK to squash then?

vext01:
I think so!

ltratt:
Squashed.

ltratt:
Hmm, that looks like a problem with rustc (which happens randomly). Although when I try manually on b13 I get a problem with rmp-serde... Investigating.

ltratt:
I think something (cargo?) has become stricter about checking something important:

```
error: failed to parse manifest at `/home/ltratt/.cargo/registry/src/github.com-1ecc6299db9ec823/rmp-serde-0.14.3/Cargo.toml`

Caused by:
  readme file with name '../README.md' was not found  
```

and indeed in its `Cargo.toml` [that's the value](https://github.com/3Hren/msgpack-rust/blob/master/rmp-serde/Cargo.toml). I assume the `../` is now treated as a security problem.

vext01:
Is there a newer rmp-serde with a fix?

vext01:
Nope...

ltratt:
You are very doubting if you think I hadn't checked :)

Here's the problem (well the fix!): https://github.com/rust-lang/cargo/commit/79c769c3d7b4c2cf6a93781575b7f592ef974255. So the next nightly will, with luck, not cause this problem.

vext01:
>  So the next nightly will, with luck, not cause this problem.

Shall we sit on this and to `bors try` over the weekend then?

ltratt:
> Shall we sit on this and to bors try over the weekend then?

I think that's the best solution (annoyingly).

smarr:
> I've only really dealt with mainstream languages.

Java: http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19
JavaScript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Unsigned_right_shift

Perhaps your mainstream languages aren't quite mainstream enough 😝

PR #149, opened 2020-06-12
Implement the 'perform:WithArguments' and friends primitives.

ltratt:
This PR implements the various `perform:withArguments` primitives.

Some of the complexity in https://github.com/softdevteam/yksom/pull/149/commits/ef599ea6e4f06de015c8036fdb82334c86b23119 is to make sure that yksom doesn't suffer from the same problems as JavaSOM (see https://github.com/SOM-st/SOM/issues/43#issuecomment-643378401). https://github.com/softdevteam/yksom/pull/149/commits/bf02eaa4da0ba4e1f0a9962a6a51ae79f2ef431b is then a simple general optimisation opened up by the first commit. I then factored out the common `perform` code in https://github.com/softdevteam/yksom/pull/149/commits/d58f3e886f761912338ae7fad4244b8c8345e5c2 which made implementing the other primitives relatively simple.

### DIFF
--- a/lang_tests/perform_in_superclass.som
+++ b/lang_tests/perform_in_superclass.som
@@ -1,0 +1,18 @@
+"
+VM:
+  status: success
+  stdout:
+    2
+    instance of perform_in_superclass
+"
+
+perform_in_superclass = (
+    run = (
+        self println.
+        self perform: #println inSuperclass: Object.
+    )
+
+    println = (
+        2 println.
+    )
+)
--- a/lang_tests/perform_in_superclass_with_args.som
+++ b/lang_tests/perform_in_superclass_with_args.som
@@ -1,0 +1,20 @@
+"
+VM:
+  status: success
+  stdout:
+    2
+    4
+    instance of perform_in_superclass_with_args
+"
+
+perform_in_superclass_with_args = (
+    run = (
+        (self ifNil: 2) println.
+        (self perform: #ifNil: withArguments: #(3) inSuperclass: Object) println.
+    )
+
+    ifNil: aBlock = (
+        aBlock println.
+        ^ 4.
+    )
+)
--- a/lang_tests/perform_witharguments.som
+++ b/lang_tests/perform_witharguments.som
@@ -1,0 +1,20 @@
+"
+VM:
+  status: success
+  stdout:
+    5
+    6
+    7
+"
+
+perform_witharguments = (
+    run = (
+        self perform: #f:b:c: withArguments: #(5 6 7).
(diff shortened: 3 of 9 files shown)


vext01:
So as far as I can see, this is like `*args` in Python? Dumb question, but the syntax is unfamiliar and it takes me quite some time to see what's going on...

smarr:
> Dumb question, but the syntax is unfamiliar and it takes me quite some time to see what's going on...

Are you referring to this?

```
self perform: #f:b:c: withArguments: #(5 6 7).
```

`#f:b:c:` is a symbol, which is essentially an interned string in Java, i.e., a string which can be compared by identity.

`#(5 6 7)` is a literal array that contains the numbers 5, 6,  and 7.

and `self perform: #f:b:c: withArguments: #(5 6 7).` combines these to a "message send" aka "method call" of the method `#perform:withArguments:`.

The `perform*` methods take the given symbol to do a lookup of a method, which is then activated given the current `self` and the arguments given in the arguments array.

So, the method name in this case is `#perform:withArguments:`, which calls a method named `#f:b:c:`.

Does that answer your question?

vext01:
Thanks for the explanation. I was wondering about some of that stuff, yeah.

I've found some decent-looking smalltalk books that I'm going to skim while stuff is compiling. I really should do my homework :)

smarr:
If you want something interactive, there's this in the browser: https://www.amber-lang.net/learn.html

This is another Smalltalk, and things can be different or missing in SOM, but, when it comes to the basic syntax, that should pretty much cover it.

The "A Little Smalltalk" book from here might be a good fit: http://stephane.ducasse.free.fr/FreeBooks.html

Again, things are somewhat different, but the gist should be compatible: http://sdmeta.gforge.inria.fr/FreeBooks/LittleSmalltalk/ALittleSmalltalk.pdf

vext01:
Thanks for the recommendations!

I was looking at:
http://stephane.ducasse.free.fr/FreeBooks/Gnu/ProgrammingUsingGnuSmalltalk.pdf

Will take a look at these too!

smarr:
The GNU Smalltalk one looks good, too. The major differences for all of these are the syntax how to define classes and methods, I think. And then differences in the libraries and set of supported literal notations. The main principles should be the same though.

ltratt:
@vext01 I'm not sure if I need to do anything else for this PR?

vext01:
> So as far as I can see, this is like *args in Python? 

It was just this. But I'm certain it's correct. So.

bors r+

PR #150, opened 2020-06-12
Update Val::div to the same format as other numeric operations.

ltratt:
This had not been modernised, which meant that it didn't handle the conversions of some types correctly.


vext01:
LGTM

bors r+

PR #151, opened 2020-06-13
An ArbInt can never be zero.

ltratt:
We ensure that is always greater than/less than an isize.

### DIFF
--- a/src/lib/vm/val.rs
+++ b/src/lib/vm/val.rs
@@ -10,7 +10,7 @@ use std::{
 
 use num_bigint::BigInt;
 use num_enum::{IntoPrimitive, UnsafeFromPrimitive};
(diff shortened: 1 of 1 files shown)


vext01:
Looks good, but it'd be good to add a comment saying why we don't handle the zero case.

You could debug assert it also?

ltratt:
It's more subtle than just the zero case: the whole integer system relies on `ArbInt`s not storing numbers that can fit in an `isize`. The comment at the top of https://raw.githubusercontent.com/softdevteam/yksom/master/src/lib/vm/objects/integers.rs hopefully explains this. The good news is that the *only* way to create an `ArbInt` is via `ArbInt::new` and it enforces this constraint.

PR #152, opened 2020-06-15
Implement instVarAt:[put:]

ltratt:
These are two relatively simple primitives, although while doing this, I realised that some existing functions should be explicitly marked as `unsafe` (https://github.com/softdevteam/yksom/commit/54422e80da444b5e22e9b2e5a6a82756d3803ae1).

### DIFF
--- a/lang_tests/inst_var_at_bad_idx.som
+++ b/lang_tests/inst_var_at_bad_idx.som
@@ -1,0 +1,17 @@
+"
+VM:
+  status: error
+  stderr:
+    Traceback...
+    ...
+    Index 2 not valid for array of length 1.
+"
+
+inst_var_at_bad_idx = (
+    | x |
+
+    run = (
+        x := 5.
+        (self instVarAt: 2) println.
+    )
+)
--- a/lang_tests/inst_var_at_put.som
+++ b/lang_tests/inst_var_at_put.som
@@ -1,0 +1,19 @@
+"
+VM:
+  status: success
+  stdout:
+    5
+    instance of inst_var_at_put
+    6
+"
+
+inst_var_at_put = (
+    | x |
+
+    run = (
+        x := 5.
+        (self instVarAt: 1) println.
+        (self instVarAt: 1 put: 6) println.
+        x println.
+    )
+)
--- a/lang_tests/inst_var_at.som
+++ b/lang_tests/inst_var_at.som
@@ -1,0 +1,14 @@
+"
+VM:
+  status: success
+  stdout: 5
+"
+
+inst_var_at = (
+    | x |
+
+    run = (
+        x := 5.
+        (self instVarAt: 1) println.
+    )
+)
--- a/lang_tests/inst_var_at_put_bad_idx.som
+++ b/lang_tests/inst_var_at_put_bad_idx.som
@@ -1,0 +1,18 @@
+"
+VM:
+  status: error
+  stderr:
+    Traceback...
+    ...
+    Index 2 not valid for array of length 1.
+"
+
+inst_var_at_put_bad_idx = (
+    | x |
+
+    run = (
+        x := 5.
+        self instVarAt: 2 put: 6.
+        x println.
+    )
+)
--- a/lang_tests/inst_var_at_put_superclass/test.som
+++ b/lang_tests/inst_var_at_put_superclass/test.som
@@ -1,0 +1,25 @@
+"
+VM:
+  status: success
+  stdout:
+    1
+    2
+    4
+    3
+"
+
+test = test_super (
+    | d c |
+
+    run = (
+        self instVarAt: 1 put: 1.
+        self instVarAt: 2 put: 2.
+        self instVarAt: 3 put: 3.
+        self instVarAt: 4 put: 4.
+
+        a println.
+        b println.
+        c println.
+        d println.
+    )
+)
--- a/lang_tests/inst_var_at_put_superclass/test_super.som
+++ b/lang_tests/inst_var_at_put_superclass/test_super.som
@@ -1,0 +1,10 @@
+test_super = (
+    | a b |
+
+    f = (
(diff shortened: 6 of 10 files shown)


lang_tests/inst_var_at_bad_idx.som
```diff
@@ -0,0 +1,17 @@
+"
+VM:
+  status: error
+  stderr:
+    Traceback...
+    ...
+    Index 2 not valid for array of length 1.
```
smarr, line 7: Hm, not the most ideal error message.

smarr:
Another test could be with a superclass that has fields.

lang_tests/inst_var_at_bad_idx.som
ltratt: Conceptually I think instance values are an array, so this seems reasonable enough to me. [I half expected them to be exposed as an `Array` as methods are!]

vext01:
> I realised that some existing functions should be explicitly marked as unsafe

Can you expand upon this? The fact that it compiled without the `unsafe` keyword suggests that -- at-least from a Rust perspective -- the functions are safe. I assume that you are using `unsafe` as a cautionary note to the programmer then?

lang_tests/inst_var_at_bad_idx.som
vext01: Wouldn't something like "instVarAt: index 2 out of range" be more user-friendly?

lang_tests/inst_var_at_bad_idx.som
```diff
@@ -0,0 +1,17 @@
...
+"
+
+inst_var_at_bad_idx = (
+    | x |
+
+    run = (
+        x := 5.
+        (self instVarAt: 2) println.
```
vext01, line 15: Ideas for more tests:

 - Getting a larger valid index (e.g. index 5 out of 7).
 - Getting a negative index should fail(?).
 - Zero index should fail(?).

lang_tests/inst_var_at_bad_idx.som
ltratt: Yes, but not worth the effort.

ltratt:
> The fact that it compiled without the unsafe keyword suggests that -- at-least from a Rust perspective -- the functions are safe.

That's not how safety in Rust works. You can package up arbitrary amounts of deadly `unsafe` code and if you don't mark the function as `unsafe`, callers will never know. `unsafe` on a function is really just a reminder to code readers that they need to be extra careful when calling it.

lang_tests/inst_var_at_bad_idx.som
ltratt: You're right... but it's just too much not-very-useful effort to do this for every indexing operation in SOM.

lang_tests/inst_var_at_bad_idx.som
vext01: OK, I'll leave that one up to you.

lang_tests/inst_var_at_bad_idx.som
vext01: It's up to you. I tend to test the edge cases, as I don't trust my ability to get it right first time ;)

lang_tests/inst_var_at_put.som
```diff
@@ -0,0 +1,19 @@
+"
+VM:
+  status: success
+  stdout:
+    5
+    instance of inst_var_at_put
```
vext01, line 6: My curiosity gets the better of me again.

This is unusual. So this is some kind of proxy? Is it transparent? Can I use arithmetic on it?

ltratt: That's just normal SOM, it's nothing really to do with this PR.

vext01: I gathered it's language semantics, was just curious.

ltratt:
@smarr's idea for a superclass test is a good one. I'll add that.

vext01:
> That's not how safety in Rust works. You can package up arbitrary amounts of deadly unsafe code and if you don't mark the function as unsafe, callers will never know. 

This doesn't match my understanding. If you include anything Rust considers unsafe in a function, [Rust will force you to mark the expression or the containing function unsafe](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e41ba8c0aecf1bc204c0e8129dd4e279).

So your annotation must be a hint to the programmer.

You could mark any function unsafe on the virtue that somewhere inside is an `unsafe` block, but by the same token you'd mark `main` unsafe.

ltratt:
> If you include anything Rust considers unsafe in a function, Rust will force you to mark the expression or the containing function unsafe.

If this was true, everything in Rust would need to be `unsafe`. `unsafe` *blocks* are the way you avoid this.

vext01:
I think we are agreeing in that aspect, but that's not what I'm questioning.

> Rust will force you to mark the expression [with an unsafe block].

As I understand it, you've added a redundant `unsafe` keyword as a programmer hint. Am I right?

I'm not saying that's wrong to do so, I've just never seen it done and I'm considering if there's utility in such an idiom.

ltratt:
Test added in https://github.com/softdevteam/yksom/pull/152/commits/89bc1ded128ae683ba81242cbdfc50548c9ef623.

vext01:
The unsafe argument was resolved in a (friendly) external discussion.

I think I buy the reasoning :P

vext01:
Unless @smarr has any further comments, I think this is OK to go in.

ltratt:
Reflecting https://github.com/SOM-st/SOM/issues/44, I've expanded the test in https://github.com/softdevteam/yksom/pull/152/commits/cf10c7ac792fada05899441ac779ef6c830d68dd (it works the same on Java SOM and yksom).

vext01:
Please squash.

ltratt:
Squashed.

PR #153, opened 2020-06-16
Implement three string primitives

ltratt:
This PR implements the `isDigits`, `isLetters`, and `isWhitespace` primitives. They're all very related. Note that I have assumed that `isDigits` only matches *ASCII* digit characters, but that the other two primitives match *Unicode* letters/whitespace.

### DIFF
--- a/lang_tests/is_whitespace.som
+++ b/lang_tests/is_whitespace.som
@@ -1,0 +1,23 @@
+"
+VM:
+  status: success
+  stdout:
+    true
+    true
+    true
+    false
+    false
+"
+
+is_whitespace = (
+    run = (
+        "Space"
+        ' ' isWhiteSpace println.
+        "Tab"
+        '	' isWhiteSpace println.
+        "Space and tab"
+        '	 ' isWhiteSpace println.
+        ' a ' isWhiteSpace println.
+        '' isWhiteSpace println.
+    )
+)
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -836,7 +836,10 @@ impl VM {
                 stry!(self.str_is(rcv, |x| x.is_ascii_digit()));
                 SendReturn::Val
             }
-            Primitive::IsLetters => unimplemented!(),
+            Primitive::IsLetters => {
+                stry!(self.str_is(rcv, |x| x.is_alphabetic()));
+                SendReturn::Val
+            }
             Primitive::IsWhiteSpace => unimplemented!(),
             Primitive::Length => {
                 // Only Arrays and Strings can have this primitive installed.
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -840,7 +840,10 @@ impl VM {
                 stry!(self.str_is(rcv, |x| x.is_alphabetic()));
                 SendReturn::Val
             }
-            Primitive::IsWhiteSpace => unimplemented!(),
+            Primitive::IsWhiteSpace => {
+                stry!(self.str_is(rcv, |x| x.is_whitespace()));
+                SendReturn::Val
+            }
             Primitive::Length => {
                 // Only Arrays and Strings can have this primitive installed.
                 debug_assert!(rcv.valkind() == ValKind::GCBOX);
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -832,7 +832,10 @@ impl VM {
             }
             Primitive::InstVarNamed => unimplemented!(),
             Primitive::InvokeOnWith => todo!(),
-            Primitive::IsDigits => unimplemented!(),
+            Primitive::IsDigits => {
+                stry!(self.str_is(rcv, |x| x.is_ascii_digit()));
+                SendReturn::Val
+            }
             Primitive::IsLetters => unimplemented!(),
             Primitive::IsWhiteSpace => unimplemented!(),
             Primitive::Length => {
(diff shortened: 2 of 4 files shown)


ltratt:
https://github.com/softdevteam/yksom/pull/153/commits/4c2aa2101b7decac58338315eccb74f8fa42d7cd handles a case(s) that I hadn't considered initially. It should probably be squashed.

src/lib/vm/core.rs
```diff
@@ -832,9 +832,36 @@ impl VM {
             }
             Primitive::InstVarNamed => unimplemented!(),
             Primitive::InvokeOnWith => todo!(),
-            Primitive::IsDigits => unimplemented!(),
-            Primitive::IsLetters => unimplemented!(),
-            Primitive::IsWhiteSpace => unimplemented!(),
+            Primitive::IsDigits => {
+                let str_val = stry!(rcv.downcast::<String_>(self));
+                let s = str_val.as_str();
+                if s.len() > 0 && s.chars().all(|x| x.is_ascii_digit()) {
```
Hirevo, line 838: Maybe using `str::is_empty` can be more idiomatic here ?
```suggestion
                if !s.is_empty() && s.chars().all(|x| x.is_ascii_digit()) {
```

src/lib/vm/core.rs
```diff
@@ -832,9 +832,36 @@ impl VM {
...
+                    self.stack.push(self.true_);
+                } else {
+                    self.stack.push(self.false_);
+                }
+                SendReturn::Val
+            }
+            Primitive::IsLetters => {
+                let str_val = stry!(rcv.downcast::<String_>(self));
+                let s = str_val.as_str();
+                if s.len() > 0 && s.chars().all(|x| x.is_alphabetic()) {
```
Hirevo, line 848: Same suggestion as above.
```suggestion
                if !s.is_empty() && s.chars().all(|x| x.is_alphabetic()) {
```

src/lib/vm/core.rs
```diff
@@ -832,9 +832,36 @@ impl VM {
...
+                    self.stack.push(self.true_);
+                } else {
+                    self.stack.push(self.false_);
+                }
+                SendReturn::Val
+            }
+            Primitive::IsWhiteSpace => {
+                let str_val = stry!(rcv.downcast::<String_>(self));
+                let s = str_val.as_str();
+                if s.len() > 0 && s.chars().all(|x| x.is_whitespace()) {
```
Hirevo, line 858: Same suggestion as above.
```suggestion
                if !s.is_empty() && s.chars().all(|x| x.is_whitespace()) {
```

smarr:
I am considering unicode support optional. Or said differently: There won't be any unicode support for CSOM or SOM++ any time soon.

Hirevo:
Oh ok, never mind my review comment then.

src/lib/vm/core.rs
ltratt: Good point!

ltratt:
The "ASCII or Unicode digits" question is an interesting one, because I wonder if all the non-ASCII numbers can be converted to integers? Maybe they can -- I don't have any realistic way of finding out that I can think of.

src/lib/vm/core.rs
ltratt: Fixed in https://github.com/softdevteam/yksom/pull/153/commits/62bf7067e48cfc834f1061d675c90ab1397c4825.

ltratt:
As an example: in Rust `is_digit` only accepts ASCII digits (but, unlike `is_ascii_digit`, `is_digit` takes a radix), whereas `is_numeric` captures the Unicode case (see e.g. https://doc.rust-lang.org/std/primitive.char.html#method.is_numeric).

lang_tests/is_whitespace.som
```diff
@@ -0,0 +1,20 @@
...
+"
+
+is_whitespace = (
+    run = (
+        "Space"
+        ' ' isWhiteSpace println.
+        "Tab"
+        '	' isWhiteSpace println.
+        ' a ' isWhiteSpace println.
+        '' isWhiteSpace println.
```
vext01, line 18: How about a test which is a mix of different whitespace? e.g. `\t \t  \t\t`.

src/lib/vm/core.rs
```diff
@@ -832,9 +832,36 @@ impl VM {
             }
             Primitive::InstVarNamed => unimplemented!(),
             Primitive::InvokeOnWith => todo!(),
-            Primitive::IsDigits => unimplemented!(),
-            Primitive::IsLetters => unimplemented!(),
-            Primitive::IsWhiteSpace => unimplemented!(),
+            Primitive::IsDigits => {
```
vext01, line 835: The three cases are really similar. Couldn't you make the code shorter by having a switch which chooses the correct closure for `.all()` and then you don't have to duplicate the rest?

vext01:
Just a couple of small comments from me.

src/lib/vm/core.rs
ltratt: I did think about this, then wasn't sure, but in retrospect you're probably right. Fixed in https://github.com/softdevteam/yksom/pull/153/commits/c7a304f6fd76163d02b829907bd5bd3597deee82.

vext01: Looks good!

lang_tests/is_whitespace.som
vext01: ^ Just this remaining.

ltratt: Fixed in https://github.com/softdevteam/yksom/pull/153/commits/6078b7ef47e0e4ec3b4416e3225c0dfbf670ef38.

vext01:
please squash

ltratt:
Squashed.

PR #154, opened 2020-06-16
Process string escapes properly.

ltratt:
Some of these are hard to test in practise (backspace and form feed are not very portable these days). The spec can only be found in the SOM test suite.

### DIFF
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -656,11 +656,38 @@ impl<'a, 'input> Compiler<'a, 'input> {
                 Ok(max_stack)
             }
             ast::Expr::String(span) => {
-                // XXX are there string escaping rules we need to take account of?
                 let s_orig = self.lexer.span_str(*span);
-                // Strip off the beginning/end quotes.
-                let s = s_orig[1..s_orig.len() - 1].to_owned();
-                let instr = Instr::String(vm.add_string(s));
+                let mut new_s = String::new();
+                // Start by ignoring the beginning quote.
+                let mut i = '\"'.len_utf8();
+                // End by ignoring the beginning quote.
+                while i < s_orig.len() - '\"'.len_utf8() {
+                    let mut c = s_orig[i..].chars().next().unwrap();
+                    if c == '\\' {
+                        i += c.len_utf8();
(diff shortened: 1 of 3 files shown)


Hirevo:
An interesting thing about string escapes in SOM that I found surprising is that, in case of a unknown escaped character (like `\a`), implementations seems to silently discard the character:
```plain
---> 'abc\adef'
it = abcdef
---> 'abc\adef' length
it = 6
```
Both the C and Java implementations simply ignore that `\a`.  
I have made som-rs to adhere to that behavior but it intrigued me when I saw it.  
I wonder if it is a strict requirement or an inefficacy that is left to implementations to decide how to handle.

smarr:
> An interesting thing about string escapes in SOM that I found surprising is that, in case of a unknown escaped character (like `\a`), implementations seems to silently discard the character

That doesn't sound like it's intended and sounds rather like a bug.
I'd probably trigger a parse error for these cases.

ltratt:
yksom throws an error in such a situation (as @smarr suggests).

vext01:
One interpretation of `\a` would be the terminal bell :P

ltratt:
AFAIK the only documentation for string escapes is here https://github.com/SOM-st/SOM/blob/master/TestSuite/StringTest.som#L94. No system bell I'm afraid ;)

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -658,9 +658,36 @@ impl<'a, 'input> Compiler<'a, 'input> {
             ast::Expr::String(span) => {
                 // XXX are there string escaping rules we need to take account of?
                 let s_orig = self.lexer.span_str(*span);
-                // Strip off the beginning/end quotes.
```
vext01, line 661: Do we need to remove the XXX two lines above too?

vext01:
Just one comment.

src/lib/compiler/ast_to_instrs.rs
ltratt: Oops, fixed in https://github.com/softdevteam/yksom/pull/154/commits/dc6f2ec5a437813e9498eedd091e9acb23483e9f.

vext01:
Please squash.

ltratt:
Squashed.

PR #155, opened 2020-06-17
Implement the 'ticks' primitive.

ltratt:


lang_tests/ticks.som
```diff
@@ -0,0 +1,11 @@
+"
+VM:
+  status: success
+  stdout: Integer
+"
+
+ticks = (
+    run = (
+        system ticks class println.
```
vext01, line 9: We could get a better test by calling it twice and checking t2 >= t1?

ltratt: Afraid not: it's not guaranteed to be monotonic...

src/lib/vm/core.rs
```diff
@@ -1021,7 +1024,22 @@ impl VM {
                 self.stack.push(v);
                 SendReturn::Val
             }
-            Primitive::Ticks => todo!(),
+            Primitive::Ticks => {
+                match SystemTime::now().duration_since(UNIX_EPOCH) {
+                    Ok(d) => {
+                        let t = d.as_micros();
+                        const_assert!(size_of::<usize>() <= size_of::<u128>());
```
vext01, line 1031: You see far into the future my friend!

lang_tests/ticks.som
ltratt: Underneath we're currently using a non-monotonic clock (it's not specified, so ...).

vext01: Ugh. OK.

ltratt: FWIW, there's not an immediately obvious way to use the monotonic timer in Rust (`Instant`) in a way that gives you seconds since the epoch. Maybe there is, maybe there isn't, but after a few minutes looking, I realised I was probably thinking about fixing a problem that isn't a problem!

src/lib/vm/core.rs
ltratt: I've been burnt before!

Hirevo:
Isn't the 'ticks' primitive supposed to give the amount of microseconds elapsed since the start of the program evaluation, instead of since the Unix epoch ?  

Consider the following program:
```smalltalk
TicksTest = (
    "Keeps the program busy for a few seconds"
    busy = (
        "You may need to adjust this value if evaluation takes too much time"
        "'som-java' takes ~5s on my machine to evaluate this"
        5000000 timesRepeat: [ nil ].
    )

    "Prints the value from `System>>#ticks` as seconds"
    printTime = (
        ('time since program boot: ' + (system ticks / 1000 / 1000) asString + ' seconds') println.
    )

    run = (
        self printTime.
        self busy.
        self printTime.
        self busy.
        self printTime.
    )
)
```

som-java prints:
```plain
time since program boot: 0 seconds
time since program boot: 5 seconds
time since program boot: 8 seconds
```
with a total execution time of around 8-9 seconds (on my machine at least).

With this PR, yksom prints a very very big number, even for the very first evaluation of `self printTime` (like: `1592414971 seconds`).

ltratt:
I'm not really sure what the difference between `time` and `ticks` is -- suggestions welcome!

Hirevo:
It seems that the only difference between `time` and `ticks` is that former is in milliseconds while the latter is in microseconds.  
At least, this is [what som-rs does](https://github.com/Hirevo/som-rs/blob/2c17a54118d28cab533bd20038d1d0bc4699922d/som-interpreter/src/primitives/system.rs#L105-L114) and what [som-java also seems to be doing](https://github.com/SOM-st/som-java/blob/09cf900a77db459624665636ee3985189d72bd1e/src/som/primitives/SystemPrimitives.java#L103-L121) in its source code.

ltratt:
I saw that, and thought my eyes must be deceiving me! @smarr, do you have any thoughts on this?

smarr:
`#ticks` is specified as `ticks = primitive      "returns the microseconds since start"` https://github.com/SOM-st/SOM/blob/master/Smalltalk/System.som#L84

Start means here indeed program start.

`#time`, well, `#time` is specified by implementation only and not really used, but yes, the main difference is the precision. `#ticks` was added after `#time`.

I think, I would expect `ticks` to be monotonic, and from a clock source useful for benchmarking.

It's specified as being from the program start to help with that where necessary.
Though, it's only really used in the context of taking a difference between two ticks.

smarr:
For the curious:

https://github.com/SOM-st/SOM/commit/a0e2c756f56cada6c367200157d395c09d644d34

```
Added convenience methods.

- this is harvested from the CSOM/PL code, contributed by students of the VM lecture at the HPI
```

ltratt:
Might this be a good opportunity to fix this? The difference is pretty confusing, and I'm not sure having two almost-but-not-quite-the-same functions is useful?

smarr:
And what's the proposed fix?

ltratt:
Deleting one of them seems the easiest fix, and defining whichever remains to be a monotonic fine-grained timer?

smarr:
Ok, I am fine with having `time` being deleted. And `ticks` being specified to return a monotonic value with microsecond resolution.

ltratt:
OK, cool. Should `ticks` then be since the start of the program or the system epoch?

smarr:
Hm, why change start of the program semantics?

ltratt:
I don't feel strongly either way.

smarr:
Since system epoch seems to encode less information.

The change comes from CSOM at a time where it was 32bit, so, the "since program start" may have been a way to squeeze useful values into 31bit integers.
I'd keep it except if there are good reasons to change it.

PR #157, opened 2020-06-17
Fix the modulus calculations.

ltratt:
I did not realise that `%` in many languages is "remainder" and that works differently than modulus for negative numbers. Lesson learned.


vext01:
I didn't know this either, but I can see how it works from experimenting in Python!

bors r+

smarr:
When I implemented these, I found interesting: https://en.wikipedia.org/wiki/Modulo_operation

PR #158, opened 2020-06-17
Fix double/int comparisons.

ltratt:
Note that these further diverge from Java SOM, where the following:

```
  (1.0 = 1) println.
  (1.0 ~= 1) println.
```

prints:

```
  true
  true
```

whereas yksom prints:

```
  true
  false
```


smarr:
Not sure, but you may run into issues with the benchmarks with this.

vext01:
Do I see it right that `1.0` is both equal and not equal to `1`?

`~=` is "not equal" as in smalltalk?

ltratt:
> Do I see it right that 1.0 is both equal and not equal to 1?

In Java SOM, yes, it's not symmetric. In yksom it is symmetric.

ltratt:
> Not sure, but you may run into issues with the benchmarks with this.

OK, I'll keep an eye open!

PR #160, opened 2020-06-17
Make fullGc return true.

ltratt:
Without this, the SOM test suite complains that this feature isn't implemented, which isn't what I expect from a test suite (but I *might* expect from a validation suite). Stefan and I disagree on whether that's reasonable or not, so this commit is a compromise: we always say we've done a full GC.

### DIFF
--- a/lang_tests/fullgc.som
+++ b/lang_tests/fullgc.som
@@ -1,7 +1,7 @@
 "
 VM:
   status: success
-  stdout: false
+  stdout: true
 "
 
 fullgc = (
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -774,7 +774,7 @@ impl VM {
                 SendReturn::Val
             }
             Primitive::FullGC => {
-                self.stack.push(self.false_);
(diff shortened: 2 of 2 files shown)


smarr:
SOMs test suite declares certain features as optional, all `expect:` tests.
You can declare me as unreasonable for doing that. I consider it pragmatic.

And I am slowly losing my energy to engage with the constant implications that what ever SOM is doing is stupid. This isn't some full time research project or industrial product. It never claimed to be perfect. It claims to be useful. Nothing more. So, I would appreciate if the tone of discussion could be a little more friendly.

smarr:
https://github.com/SOM-st/SOM/issues/48

ltratt:
If https://github.com/SOM-st/SOM/issues/48 goes ahead, then I think there is no need for this PR.

vext01:
> If SOM-st/SOM#48 goes ahead, then I think there is no need for this PR.

OK, I'll hold off reviewing this for now.

ltratt:
https://github.com/SOM-st/SOM/pull/49 has been merged, so we no longre need this PR.

PR #163, opened 2020-06-29
Run the SOM testsuite on buildbot.

ltratt:
Since https://github.com/SOM-st/SOM/pull/49 has been merged, we can now easily run the test suite, so I think it makes sense to do that during CI.


ltratt:
bors try

PR #164, opened 2020-06-29
We don't use specialisation.

ltratt:
Recent rustc nightlies warn against using this feature so, since we don't need it, we might as well remove it.


vext01:
I've never used specialisation, but if we don't use it, then great.

bors r+

PR #167, opened 2020-07-07
Use SmartString.

ltratt:
The [SmartString](https://crates.io/crates/smartstring) library avoids allocating memory for strings if they're small enough to fit inside the existing String struct. This is quite appealing for an interpreter where in general there are quite a lot of small strings that are
critical for performance (e.g. function names; though, admittedly, Smalltalk has many unusually long function names).

This commit is a fairly simple port (note that we generally import `SmartString` as `String` i.e. `String` no longer means `std::string::String` as per normal). Here's how long the test suite takes on my desktop before this commit:

```
  $ multitime -n 30 -s0 -q target/release/yksom --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
  ===> multitime results
  1: -q target/release/yksom --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
              Mean        Std.Dev.    Min         Median      Max
  real        0.214       0.002       0.210       0.214       0.220
  user        0.318       0.028       0.260       0.310       0.360
  sys         0.015       0.011       0.000       0.015       0.040
```

And here's after:

```
  $ multitime -n 30 -s0 -q target/release/yksom --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
  ===> multitime results
  1: -q target/release/yksom --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
              Mean        Std.Dev.    Min         Median      Max
  real        0.194       0.004       0.189       0.193       0.207
  user        0.259       0.035       0.200       0.260       0.310
  sys         0.015       0.013       0.000       0.010       0.050
```

The mean time has decreased by about 10% which is good, but the big question is whether this is just due to our currently-very-slow GC or not. In other words, would this performance change decrease (almost certainly) or disappear (who knows) if/when our GC performance improves. This is something of an unknowable at the moment and means that I'm unsure whether we should consider this for merging or not.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ num-integer = "0.1"
 num_enum = "0.5"
 num-traits = "0.2"
 ryu = "1.0"
+smartstring = "0.2"
 static_assertions = "1.1"
 termion = "1.5"
 
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -7,6 +7,7 @@ use std::{
 use itertools::Itertools;
 use lrpar::{NonStreamingLexer, Span};
 use rboehm::Gc;
+use smartstring::alias::String as SmartString;
 
 use crate::{
     compiler::{
@@ -25,7 +26,7 @@ pub struct Compiler<'a, 'input> {
     lexer: &'a dyn NonStreamingLexer<'input, StorageT>,
     path: &'a Path,
     /// The stack of variables at the current point of evaluation.
-    vars_stack: Vec<HashMap<String, usize>>,
+    vars_stack: Vec<HashMap<SmartString, usize>>,
     /// Since SOM's "^" operator returns from the enclosed method, we need to track whether we are
     /// in a closure -- and, if so, how many nested closures we are inside at the current point of
     /// evaluation.
@@ -40,7 +41,7 @@ impl<'a, 'input> Compiler<'a, 'input> {
         lexer: &dyn NonStreamingLexer<StorageT>,
         path: &Path,
         astcls: &ast::Class,
-    ) -> Result<(String, Val), String> {
+    ) -> Result<(SmartString, Val), String> {
         let mut compiler = Compiler {
             lexer,
             path,
@@ -48,8 +49,8 @@ impl<'a, 'input> Compiler<'a, 'input> {
             closure_depth: 0,
         };
 
-        let name = lexer.span_str(astcls.name).to_owned();
-        let (supercls_val, supercls_meta_val) = if name != "Object" {
+        let name = SmartString::from(lexer.span_str(astcls.name));
+        let (supercls_val, supercls_meta_val) = if name.as_str() != "Object" {
             let supercls_val = if let Some(n) = astcls.supername.map(|x| lexer.span_str(x)) {
                 match vm.load_class(n) {
                     Ok(cls) => cls,
@@ -84,7 +85,7 @@ impl<'a, 'input> Compiler<'a, 'input> {
         let metacls = match compiler.c_class(
             vm,
             lexer,
-            format!("{} class", &name),
+            format!("{} class", name.as_str()).into(),
             supercls_meta_val,
             &astcls.class_inst_vars,
             &astcls.class_methods,
@@ -137,7 +138,7 @@ impl<'a, 'input> Compiler<'a, 'input> {
         &mut self,
         vm: &mut VM,
         lexer: &'a dyn NonStreamingLexer<StorageT>,
-        name: String,
+        name: SmartString,
         supercls_val: Val,
         ast_inst_vars: &[Span],
         ast_methods: &[ast::Method],
@@ -168,7 +169,7 @@ impl<'a, 'input> Compiler<'a, 'input> {
         };
         for var in ast_inst_vars {
             let vars_len = inst_vars.len();
-            let n = lexer.span_str(*var).to_owned();
+            let n = SmartString::from(lexer.span_str(*var));
             match inst_vars.entry(n) {
                 hash_map::Entry::Occupied(_) => {
                     return Err(vec![(
@@ -203,8 +204,8 @@ impl<'a, 'input> Compiler<'a, 'input> {
         }
         let inst_vars_map = self.vars_stack[0]
             .iter()
-            .map(|(k, v)| (k.to_string(), *v))
-            .collect::<HashMap<String, usize>>();
+            .map(|(k, v)| (k.clone(), *v))
+            .collect::<HashMap<SmartString, usize>>();
         self.vars_stack.pop();
         if !errs.is_empty() {
             return Err(errs);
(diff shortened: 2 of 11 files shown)


smarr:
Is it an option to disable GC and see what happens, i.e., run without any memory management?
I'd also be curious what the impact on the Json benchmark is (in case that one runs already).
[JsonSmall](https://github.com/SOM-st/SOM/blob/master/Examples/Benchmarks/Json/JsonSmall.som) should work on interpreters.

ltratt:
> Is it an option to disable GC and see what happens, i.e., run without any memory management?

I'm not sure. @jacob-hughes might know. Though blowing up memory usage comes with its own problems!

vext01:
In terms of code changes, looks OK to me.

Perhaps I'd have not shadowed `std::String`, but I won't make you go back and update all of the places.

In terms of performance. I was intriuged to see what difference smart strings would make.

Here's a program which makes short strings repeatedly. An assertion prevents the compiler from removing the strings altogether:
```
//use smartstring::alias::String;
use std::fmt::Write;

fn main() {
    for i in 0..10000000 {
        let mut s = String::new();
        write!(&mut s, "i{}", i).unwrap();
        assert!(s.len() <= 9);
    }
}
```

Without smart strings:
```
$ multitime -n 10 ./target/release/smartstring_perf
===> multitime results
1: ./target/release/smartstring_perf
            Mean        Std.Dev.    Min         Median      Max
real        2.240       0.008       2.233       2.238       2.263       
user        2.234       0.007       2.220       2.235       2.240       
sys         0.000       0.000       0.000       0.000       0.000       
```

With smart strings:
```
$ multitime -n 10 ./target/release/smartstring_perf
===> multitime results
1: ./target/release/smartstring_perf
            Mean        Std.Dev.    Min         Median      Max
real        0.481       0.002       0.479       0.480       0.487       
user        0.479       0.005       0.470       0.480       0.490       
sys         0.002       0.004       0.000       0.000       0.010       
```

And now repeating this, using long strings, that (hopefully) can't be inlined:
```
//use smartstring::alias::String;
use std::fmt::Write;

const LONG: &str = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";

fn main() {
    for i in 0..10000000 {
        let mut s = String::new();
        write!(&mut s, "{}{}", LONG, i).unwrap();
        assert!(s.len() <= 1000);
    }
}
```

(This should be a worst case for smart strings, as the initial empty string will be inlined, but then will be demoted to the heap at `write!`).

Without smart strings:
```
$ multitime -n 10 ./target/release/smartstring_perf
===> multitime results
1: ./target/release/smartstring_perf
            Mean        Std.Dev.    Min         Median      Max
real        4.537       0.031       4.495       4.524       4.597       
user        4.526       0.039       4.470       4.520       4.600       
sys         0.002       0.004       0.000       0.000       0.010       
```

And with them:
```
$ multitime -n 10 ./target/release/smartstring_perf
===> multitime results
1: ./target/release/smartstring_perf
            Mean        Std.Dev.    Min         Median      Max
real        4.564       0.031       4.523       4.556       4.608       
user        4.555       0.037       4.510       4.545       4.610       
sys         0.003       0.005       0.000       0.000       0.010       
```

So based on this, I'd say smart strings are a net gain?

ltratt:
Smart strings are *probably* a gain, but I'd really want to be a bit more sure before we merge a PR like this that makes the code less clear. For all I know, string allocations are such a small part of yksom that it's not worth muddying the code, and the gains I saw are just because our GC is currently very slow.

vext01:
FWIW, I don't think it muddies the code too much.

Do we know the threshold for smart string inlining?

If so, can we write a destructor for our SOM string class that prints to stderr if the string was inlined or not? Then we can see how many strings are likely to be a performance win.

ltratt:
> Do we know the threshold for smart string inlining?

Yes, it's documented (though, of course, it changes depending on the word size of the machine!).

> If so, can we write a destructor for our SOM string class that prints to stderr if the string was inlined or not?

Note that we smartstring-ise nearly *all* strings in yksom, including those that aren't SOM strings. For example, method lookups (which previously were normal Rust `String`s) are smartstrings in this PR. So it's a bit difficult to get accurate numbers.

vext01:
Would you feel better if we didn't shadow `std::String` so that the use-sites are explicit? Then this doesn't muddy the code so much.

jacob-hughes:
> > Is it an option to disable GC and see what happens, i.e., run without any memory management?
> 
> I'm not sure. @jacob-hughes might know. Though blowing up memory usage comes with its own problems!

I think you should be able to run yksom with the `GC_DONT_GC=true` environment variable.

jacob-hughes:
This is a cool idea. I'd not heard of this library. LGTM and results seem promising, even if, as you say, they're amplified by just how poor the GC performs right now.

ltratt:
> Would you feel better if we didn't shadow std::String so that the use-sites are explicit? 

Yes, on reflection I would. Fixed in https://github.com/softdevteam/yksom/pull/167/commits/8ad8fc5eb5ac9056ad708af108384abfc7436895.

ltratt:
> I think you should be able to run yksom with the GC_DONT_GC=true environment variable.

Aha. I didn't know that.

Before this PR:

```
ltratt@overdrive GC_DONT_GC=true multitime -n 30 -s0 -q target/release/yksomo --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
===> multitime results
1: -q target/release/yksomo --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
            Mean        Std.Dev.    Min         Median      Max
real        0.111       0.002       0.109       0.111       0.118       
user        0.095       0.009       0.080       0.100       0.110       
sys         0.008       0.010       0.000       0.005       0.030  
```

With this PR:

```
GC_DONT_GC=true multitime -n 30 -s0 -q target/release/yksom --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
===> multitime results
1: -q target/release/yksom --cp SOM/Smalltalk/ SOM/TestSuite/TestHarness.som
            Mean        Std.Dev.    Min         Median      Max
real        0.112       0.003       0.109       0.111       0.122       
user        0.092       0.011       0.070       0.090       0.110       
sys         0.013       0.010       0.000       0.010       0.040       
```

So for this example, we're not making any observable difference.

I think I need to see if I can get some SOM benchmarks running. They're not immediately running, but the fix may be easy (or not...).

vext01:
I prefer this too.

So are we going ahead with this? If so, please squash.

ltratt:
> So are we going ahead with this? If so, please squash.

Let me see if I can get a benchmark or two running first.

ltratt:
https://github.com/softdevteam/yksom/pull/168 allows us to run the JSON benchmark (and maybe others, but I haven't tried those yet). Before this PR:

```
1: target/release/yksom --cp SOM/Smalltalk --cp SOM/Examples/Benchmarks SOM/Examples/Benchmarks/Json/Json.som
            Mean        Std.Dev.    Min         Median      Max
real        58.799      0.364       58.336      58.835      59.227      
user        97.090      0.477       96.420      97.360      97.490      
sys         0.533       0.175       0.300       0.580       0.720   
```

With this PR:

```
1: target/release/yksom --cp SOM/Smalltalk --cp SOM/Examples/Benchmarks SOM/Examples/Benchmarks/Json/Json.som
            Mean        Std.Dev.    Min         Median      Max
real        54.711      0.591       53.901      54.938      55.295      
user        76.453      0.638       75.590      76.660      77.110      
sys         0.510       0.098       0.390       0.510       0.630    
```

So that's a useful improvement. Note that `user` is bigger than `real`: you can see Boehm doing a lot of work in the second thread in `top`!

The numbers above are from my desktop. I don't have enough RAM to run without GC there but I do on bencher13 where, with `GC_DONT_GC=true` before this PR:

```
1: target/release/yksom --cp SOM/Smalltalk --cp SOM/Examples/Benchmarks SOM/Examples/Benchmarks/Json/Json.som
            Mean        Std.Dev.    Min         Median      Max
real        19.432      0.047       19.399      19.399      19.498      
user        18.196      0.050       18.160      18.162      18.267      
sys         1.237       0.003       1.232       1.238       1.240   
```

With this PR:

```
1: target/release/yksom --cp SOM/Smalltalk --cp SOM/Examples/Benchmarks SOM/Examples/Benchmarks/Json/Json.som
            Mean        Std.Dev.    Min         Median      Max
real        19.152      0.037       19.100      19.170      19.185      
user        17.933      0.092       17.817      17.942      18.040      
sys         1.220       0.064       1.132       1.244       1.284  
```

So, even with the GC turned fully off, this PR does bring a small improvement. I think we should merge https://github.com/softdevteam/yksom/pull/168 first, then I can rebase this against master and we can merge.

vext01:
Do you want to do the rebase now?

ltratt:
Rebased and squashed.

PR #168, opened 2020-07-08
Only the first receiver in message sends should be a super send.

ltratt:
See https://github.com/SOM-st/SOM/issues/53 and https://github.com/SOM-st/SOM/pull/31.


vext01:
Makes sense to me.

bors r+

PR #170, opened 2020-07-08
Clarify the layers of instance variables.

ltratt:
Because of SOM's meta-class hierarchy, it's quite easy to get a bit lost as to where you are in it. I compounded that by not understanding it at first and hacking it in a bit later. That meant that we had a variable "num_inst_vars" in classes which described how many instance variables *instances* of the class would possess, but not how many the class itself has. Since I just managed to thoroughly confuse myself over this, this commit moves a few things around so that we no longer have that attribute, which hopefully makes things a little clearer for the next person who stumbles into this part of the code.


vext01:
OK. My review is superficial, as I don't know about this meta-class hierarchy.

Code-wise, looks fine.

bors r+

PR #171, opened 2020-07-23
Use arbitrary self types so that we can use `self: Gc<Self>`.

ltratt:
VM objects are all allocated on the heap. The Val::recover function turns an object back into a `Val`, but is deeply unsafe: if you didn't allocate the object via `Val::from_obj` all sorts of bad, weird things will happen. Previously we entirely relied on the programmer doing The Right Thing in order that the bad case didn't occur.

This commit makes it *much* clearer that objects are `Gc`d things by making their `self` types `self: Gc<Self>`. For example this means that String::concatenate function previously looked like this:

```rust
pub fn concatenate(&self, ...) -> Result<Val, Box<VMError>> {
    ...
    Val::recover(self)
    ...
}
```

But you could call that with a non-GCd `&self` and then things would explode. This function now looks like:

```rust
pub fn concatenate(self: Gc<Self>, ...) -> Result<Val, Box<VMError>> {
    ...
    Val::recover(self)
    ...
}
```

and `Val::recover` now looks like:

```rust
pub fn recover<T: Obj + 'static>(obj: Gc<T>) -> Self {
```

so we can statically guarantee that you can't `recover` something that wasn't Gc'd.rs

There is more to do to squeeze more static guarantees out of the codebase, but this is the minimal change which starts that process off.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -9,20 +9,13 @@ export CARGO_HOME="`pwd`/.cargo"
 export RUSTUP_HOME="`pwd`/.rustup"
 
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh
-sh rustup.sh --default-host x86_64-unknown-linux-gnu --default-toolchain nightly -y --no-modify-path
+# Install no toolchain initially to ensure toolchain rollback.
+sh rustup.sh --default-host x86_64-unknown-linux-gnu --default-toolchain none -y --no-modify-path
 
 export PATH=`pwd`/.cargo/bin/:$PATH
+rustup install nightly
 
-# Sometimes rustfmt is so broken that there is no way to install it at all.
-# Rather than refusing to merge, we just can't rust rustfmt at such a point.
-rustup component add --toolchain nightly rustfmt-preview \
-    || cargo +nightly install --force rustfmt-nightly \
-    || true
-rustfmt=0
-cargo fmt 2>&1 | grep "not installed for the toolchain" > /dev/null || rustfmt=1
-if [ $rustfmt -eq 1 ]; then
-    cargo +nightly fmt --all -- --check
-fi
+cargo +nightly fmt --all -- --check
 
 cargo test
 cargo test --release
--- a/src/lib/mod.rs
+++ b/src/lib/mod.rs
@@ -12,8 +12,10 @@
 
 #![feature(alloc_layout_extra)]
 #![feature(allocator_api)]
+#![feature(arbitrary_self_types)]
 #![feature(box_patterns)]
 #![feature(coerce_unsized)]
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -31,14 +31,16 @@ lazy_static = "1.3"
 regex = "1.1"
 
 [dependencies]
-rboehm = { git = "https://github.com/softdevteam/rboehm" }
+#rboehm = { git = "https://github.com/softdevteam/rboehm" }
+rboehm = { path = "../rboehm" }
```
jacob-hughes, line 35: This will need changing back in order for this PR to be mergeable

Cargo.toml
```diff
@@ -31,14 +31,16 @@ lazy_static = "1.3"
...
 arrayvec = "0.5"
 cfgrammar = "0.9"
 getopts = "0.2"
 itertools = "0.9"
 lrlex = "0.9"
 lrpar = "0.9"
-natrob = { git="https://github.com/softdevteam/natrob", features=["rboehm"] }
+#natrob = { git="https://github.com/softdevteam/natrob", features=["rboehm"] }
+natrob = { path="../natrob", features=["rboehm"] }
```
jacob-hughes, line 43: And here

Cargo.toml
ltratt: Agreed: this is very much a draft (though I should still have not committed these two lines!) while we work out if this is a good idea or not.

ltratt:
@jacob-hughes I think I'm still happy with this PR, although that last commit will need squashing!

ltratt:
[I did a quick JSON benchmark and there's no statistically significant performance change that I can see.]

jacob-hughes:
> [I did a quick JSON benchmark and there's no statistically significant performance change that I can see.]

Cool. I think that's a good thing. I didn't expect any performance improvement from this but no regression is always nice.

ltratt:
OK for me to squash?

jacob-hughes:
Please go ahead!

ltratt:
Squashed.

jacob-hughes:
A small nit, shall we change the commit title to:

"Use arbitrary self types for VM objects."

ltratt:
Shortened (not quite as suggested... but still shorter!).

jacob-hughes:
Do you want to change that in the commit message too?

ltratt:
Forced pushed an updated commit title.

jacob-hughes:
LGTM

bors r+

ltratt:
Gah, rustfmt. Fix coming.

ltratt:
Please try that.

jacob-hughes:
Looks good. Please squash

ltratt:
I think that new commit is best left as-is.

jacob-hughes:
Cool.

bors r+

PR #175, opened 2020-08-06
Split `Frame::new` into two.

ltratt:
This function is one of the most frequently called in yksom, so the fact that it serves two different masters (methods and blocks) is less than ideal (at the very least it introduces a branch). This commit splits the function into two, one for methods and one for blocks.

To some extent this is a bit pointless in that the major overhead of `Frame::new` is the malloc of a `Vec`, but one day that cost will be much less, so this change seems like a useful one, if perhaps a bit early!

This can be reviewed but won't be mergeable until https://github.com/softdevteam/rboehm/pull/19 is merged.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -341,7 +341,7 @@ impl VM {
                 for a in args {
                     self.stack.push(a);
                 }
-                let frame = Frame::new(self, None, rcv, None, num_vars, meth.num_params());
+                let frame = Frame::new_method(self, rcv, num_vars, meth.num_params());
                 self.frames.push(frame);
                 let r = self.exec_user(rcv, meth, bytecode_off);
                 self.frame_pop();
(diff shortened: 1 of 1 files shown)


src/lib/vm/core.rs
```diff
@@ -341,7 +341,7 @@ impl VM {
                 for a in args {
                     self.stack.push(a);
                 }
-                let frame = Frame::new(self, None, rcv, None, num_vars, meth.num_params());
+                let frame = Frame::new_method(self, rcv, num_vars, meth.num_params());
```
smarr, line 344: This reads a little strange. If this isn't a Rust-ism, I'd wonder whether `Frame::new_for_method`/``Frame::new_for_block` or something like it could be just a little more clear.

ltratt: It's a bit tricky because `from_*` has an idiomatic meaning in Rust (which wouldn't really make sense in this case)... Names are hard :/

smarr: `from_`?  Do you mean `for`? Other option `new_with_method`

ptersilie:
Looks good.

PR #177, opened 2020-08-06
Remove calls to tobj() when we know we must have a GCBOX.

ltratt:
Needs https://github.com/softdevteam/rboehm/pull/19 to be merged first.

At the moment a `Val` can only be `INT` or `GCBOX`, so in cases where we've checked to see if `self` is an `INT`, the only remaining case is `GCBOX`. The locations changed in this commit implicitly checked that `self` was a `GCBOX`
twice: by using unsafe `gcbox_to_tobj` function, we can remove the second (pointless) check. This is a just about measurable improvement on the JSON benchmark at about 1%.

### DIFF
--- a/src/lib/vm/val.rs
+++ b/src/lib/vm/val.rs
@@ -344,7 +344,8 @@ impl Val {
                 VMErrorKind::BuiltinTypeError { expected, got },
             ));
         }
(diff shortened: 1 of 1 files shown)


ptersilie:
Looks good.

PR #172, opened 2020-07-25
Create variable sized objects

ltratt:
**Draft PR: we need to check that we're confident this really does what we think it should do, as it uses a new-ish Rust feature.**

Previously we created objects on the stack and then moved them into the heap. Mostly this is cheap (and most of the moves were probably optimised away), but it is clearly inefficient for large objects, and it also limits us to objects that are a fixed size.

This commit changes the object creation API substantially. Objects now request a chunk of heap via `Val::heap_obj`, to which they are handed a raw pointer, to which they must write a fully initialised object. We use a function for this, so that the bounds of uninitialisation are clear. For example previously we wrote:

```rust
    pub fn from_vec(vm: &mut VM, store: Vec<Val>) -> Val {
        Val::from_obj(
            vm,
            NormalArray {
                store: UnsafeCell::new(store),
            }
        }
    }
```

which now becomes:

```rust
    pub fn from_vec(_: &mut VM, store: Vec<Val>) -> Val {
        Val::heap_obj(
            Layout::new::<NormalArray>(),
            |ptr: *mut NormalArray| unsafe {
                let ptr = &raw mut *ptr;
                (*ptr).store = UnsafeCell::new(store);
            },
        )
    }
```

As this shows we use the new "raw reference" feature (see https://github.com/rust-lang/rfcs/pull/2582/files) to make this easier to use. [In an ideal world, the init function would take an `&raw mut *T` but that doesn't seem to be supported yet, so each init function has to do the appropriate type conversion itself.]

It is important to note that the init function *must* fully initialise the memory for the main object; however, if extra memory is requested (which we don't do yet... but will soon!), then that does not need to be initialised.

[Note that, for some reason that may once have made sense, `Class`es and `Method`s did not return `Val`s even though the functions that created them immediately converted them to `Val`s; this commit changes them too.]

There isn't any observable speed change from this commit, but it will hopefully open up opportunities for speed improvements later.

PR #179, opened 2020-08-11
Normal array fewer allocations

ltratt:
This PR makes `NormalArray` (fixed sized arrays in SOM) perform a single allocation (similar to how `Inst` is stored after #172), and also changes a couple of cases to not allocate intermediate `Vec`s.

### DIFF
--- a/src/lib/vm/objects/array.rs
+++ b/src/lib/vm/objects/array.rs
@@ -137,7 +137,7 @@ impl Array for NormalArray {
 impl NormalArray {
     pub fn new(vm: &VM, len: usize) -> Val {
         unsafe {
-            NormalArray::alloc(vm, len, |mut store: *mut Val| {
+            NormalArray::alloc(len, |mut store: *mut Val| {
                 for _ in 0..len {
                     store.write(vm.nil);
                     store = store.add(1);
@@ -146,9 +146,9 @@ impl NormalArray {
         }
     }
 
-    pub fn from_vec(vm: &VM, v: Vec<Val>) -> Val {
+    pub fn from_vec(v: Vec<Val>) -> Val {
         unsafe {
-            NormalArray::alloc(vm, v.len(), |store: *mut Val| {
+            NormalArray::alloc(v.len(), |store: *mut Val| {
                 copy_nonoverlapping(v.as_ptr(), store, v.len());
             })
         }
@@ -158,7 +158,7 @@ impl NormalArray {
     /// with a pointer to the uninitialised backing store. After `init` completes, the backing
     /// store will be considered fully initialised: failure to fully initialise it causes undefined
     /// behaviour.
-    unsafe fn alloc<F>(_: &VM, len: usize, init: F) -> Val
+    unsafe fn alloc<F>(len: usize, init: F) -> Val
     where
         F: FnOnce(*mut Val),
     {
--- a/src/lib/vm/objects/array.rs
+++ b/src/lib/vm/objects/array.rs
@@ -1,6 +1,9 @@
 #![allow(clippy::new_ret_no_self)]
 
-use std::{cell::UnsafeCell, collections::hash_map::DefaultHasher, hash::Hasher};
+use std::{
+    alloc::Layout, cell::UnsafeCell, collections::hash_map::DefaultHasher, hash::Hasher,
+    mem::size_of, ptr::copy_nonoverlapping,
+};
 
 use rboehm::Gc;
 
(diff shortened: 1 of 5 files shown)


src/lib/vm/objects/array.rs
```diff
@@ -30,7 +33,27 @@ pub trait Array {
...
+}
+
+// Since arrays have a fixed number of elements in their store, we do not need to allocate a separate
+// object and store: we can fit both in a single block. We thus use a custom layout where
+// the `NormalArray` comes first, followed immediately by a contiguous array of `Val`s. On a
+// 64-bit machine this looks roughly as follows:
+//
+//   0: NormalArray { len: ... }
+//   8: Val [representing instance field 0]
+//   16: Val [representing instance field 1]
```
vext01, line 46: I was about to ask: why would an array contain an arbitrary number of arrays, but I think I now understand.

What's confusuing about this doc string is the use of `[]`. I thought that the comments in these braces were arrays themselves. I think it'd be clearer as:

```
//   0: NormalArray { len: ... }
//   8: Val_0
//   16: Val_1
//    ...
//    _: Val_{len - 1}
```

Or similar?

src/lib/vm/objects/array.rs
```diff
@@ -67,42 +89,37 @@ impl StaticObjType for NormalArray {
...
         }
     }
 
-    unsafe fn unchecked_at(self: Gc<Self>, mut idx: usize) -> Val {
-        debug_assert!(idx > 0);
-        let store = &*self.store.get();
-        debug_assert!(idx <= store.len());
-        idx -= 1;
-        *store.get_unchecked(idx)
+    unsafe fn unchecked_at(self: Gc<Self>, idx: usize) -> Val {
```
vext01, line 106: I think some of these functions could do with a comment saying what they do (but then again I'm not well-versed at the insides of yksom).

src/lib/vm/objects/array.rs
```diff
@@ -111,24 +128,52 @@ impl Array for NormalArray {
...
+    pub fn from_vec(vm: &VM, v: Vec<Val>) -> Val {
+        unsafe {
+            NormalArray::alloc(vm, v.len(), |store: *mut Val| {
+                copy_nonoverlapping(v.as_ptr(), store, v.len());
+            })
+        }
+    }
+
+    /// Create a new `NormalArray` with a backing store of `len` elements. `init` will be called
+    /// with a pointer to the uninitialised backing store. After `init` completes, the backing
```
vext01, line 158: You are saying: you must call `init()` after this?

Perhaps add "Only"?:
```
Only after init completes...
```

src/lib/vm/objects/array.rs
```diff
@@ -111,24 +128,52 @@ impl Array for NormalArray {
...
+    /// behaviour.
+    unsafe fn alloc<F>(_: &VM, len: usize, init: F) -> Val
+    where
+        F: FnOnce(*mut Val),
+    {
+        let (vals_layout, vals_dist) = Layout::new::<Val>().repeat(len).unwrap();
+        // We require the store to be laid out as a C-like contiguous array.
+        debug_assert_eq!(vals_dist, size_of::<Val>());
+        let (layout, store_off) = Layout::new::<NormalArray>().extend(vals_layout).unwrap();
+        // We require the store to be positioned immediately after the `NormalArray` with no
```
vext01, line 169: Will the lack of padding ever cause alignment issues?

src/lib/vm/objects/array.rs
ltratt: They are documented in the trait they implement (which is the Rust convention for these things AFAICT).

src/lib/vm/somstack.rs
```diff
@@ -75,19 +75,17 @@ impl SOMStack {
 
     /// Splits the collection into two at the given index.
     ///
-    /// Returns a newly allocated NormalArray containing the elements in the range [at, len). After
-    /// the call, the SOM stack will be left containing the elements [0, at) with its previous
-    /// capacity unchanged.
+    /// Returns a newly allocated `NormalArray` containing the elements in the range [at, len).
+    /// After the call, the SOM stack will be left containing the elements [0, at) with its
```
vext01, line 79: It has been such a long time since I've seen interval syntax that I had to look it up.

Since this is a rust program, I wonder if it'd be clearer to write:
```
will be left containing the elements `0..at`
```

i.e. using familiar range syntax.

Total bikeshedding of course.

vext01:
A somewhat shallow review. Handful of comments.

src/lib/vm/objects/array.rs
vext01: Ah OK. My bad.

src/lib/vm/somstack.rs
ltratt: I think that comment is adapted from the Rust std library so is best left as-is (I agree that the syntax is unfamiliar).

vext01: fair enough

src/lib/vm/objects/array.rs
ltratt: `init` is a closure/function you pass in, so adding "only" wouldn't be correct in this case.

vext01: Gotcha.

src/lib/vm/objects/array.rs
ltratt: If it does, this assignment will fire. [In practise I think it's rather unlikely to, but this is me being paranoid!]

src/lib/vm/objects/array.rs
ltratt: Point taken. Fixed in https://github.com/softdevteam/yksom/pull/179/commits/cf068f8539767133e6b3b505f799edf6f518c583.

src/lib/vm/objects/array.rs
vext01: ok

vext01:
Please squash

ltratt:
Squashed.

PR #184, opened 2020-09-04
Store SOM variables on the SOM stack.

ltratt:
Previously each function call allocated memory for variables on the heap. This commit moves things so that most variables live and die on the SOM stack. This then enables doing many more operations on the stack, including passing function arguments in on the stack rather than through intermediate `Vec`s (see the comment in `somtack.rs`). This is a substantial speed-up (a little under 2x).

The scheme we use is, more-or-less, that of Lua's. The most approachable explanation I know of is:

  http://www.craftinginterpreters.com/closures.html

though it's not something I'd expect anyone to understand in 5 minutes. The basic idea is that methods (and blocks which don't reference variables in outer methods / closures) don't require any heap allocation of variables. Blocks which reference variables in outer methods / closures pre-allocate memory on the heap and, if necessary, move variables into that memory when the outer method / block completes. More sophisticated schemes are possible, but this scheme has the advantage of relative simplicity.

Note that there is a known case of UB in this code (which Jake and I are musing how best to solve in the longer term): we currently don't have a way of sensibly allowing mutation in interior pointers to `Gc` things, which makes the `UpVar` struct dangerous. In practise, at the moment, I can't observe the UB causing us problems, and I think we'll have a fix for it before it becomes a bigger problem. I think, for the time being, it's probably easiest to review this commit as if that wasn't an issue, as the enabling fix will have to happen in external libraries.

### DIFF
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -12,7 +12,7 @@ use smartstring::alias::String as SmartString;
 use crate::{
     compiler::{
         ast,
-        instrs::{Instr, Primitive},
+        instrs::{Instr, Primitive, UpVarDef},
         StorageT,
     },
     vm::{
@@ -23,11 +23,16 @@ use crate::{
     },
 };
 
+// A variable with a name which the user cannot reference.
+const CLOSURE_RETURN_VAR: &str = "$$closurereturn$$";
+
 pub struct Compiler<'a, 'input> {
     lexer: &'a dyn NonStreamingLexer<'input, StorageT>,
     path: &'a Path,
-    /// The stack of variables at the current point of evaluation.
-    vars_stack: Vec<HashMap<SmartString, usize>>,
+    upvars_stack: Vec<Vec<(SmartString, UpVarDef)>>,
+    /// The stack of local variables at the current point of evaluation. Maps variable names to
+    /// `(captured, var idx)`.
+    locals_stack: Vec<HashMap<SmartString, (bool, usize)>>,
     blocks_stack: Vec<Vec<Gc<Function>>>,
     /// Since SOM's "^" operator returns from the enclosed method, we need to track whether we are
     /// in a closure -- and, if so, how many nested closures we are inside at the current point of
@@ -36,6 +41,14 @@ pub struct Compiler<'a, 'input> {
 }
 
 type CompileResult<T> = Result<T, Vec<(Span, String)>>;
+
+#[derive(Debug)]
+enum VarLookup {
+    Local(usize),
+    UpVar(usize),
+    InstVar(usize),
+    Unknown,
+}
 
 impl<'a, 'input> Compiler<'a, 'input> {
     pub fn compile(
@@ -47,7 +60,8 @@ impl<'a, 'input> Compiler<'a, 'input> {
         let mut compiler = Compiler {
             lexer,
             path,
-            vars_stack: Vec::new(),
+            locals_stack: Vec::new(),
+            upvars_stack: Vec::new(),
             blocks_stack: Vec::new(),
             closure_depth: 0,
         };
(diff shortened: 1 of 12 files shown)


ltratt:
bors try

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -23,11 +23,16 @@ use crate::{
...
+// A variable with a name which the user cannot reference.
+const CLOSURE_RETURN_VAR: &str = "$$closurereturn$$";
+
 pub struct Compiler<'a, 'input> {
     lexer: &'a dyn NonStreamingLexer<'input, StorageT>,
     path: &'a Path,
-    /// The stack of variables at the current point of evaluation.
-    vars_stack: Vec<HashMap<SmartString, usize>>,
+    upvars_stack: Vec<Vec<(SmartString, UpVarDef)>>,
+    /// The stack of local variables at the current point of evaluation. Maps to `(captured, var
```
vext01, line 33: So the key for this map (and the upvars map) is a variable name? Can we add/tweak comments to make it clear?

ltratt: Fixed in https://github.com/softdevteam/yksom/pull/184/commits/85e79df8c4ab8a00c130236a523706c95ca67842.

ltratt: Fixed in https://github.com/softdevteam/yksom/pull/184/commits/85e79df8c4ab8a00c130236a523706c95ca67842.

vext01: Did we want to do the same for upvars, or do we feel it's obvious?

src/lib/vm/core.rs
```diff
@@ -411,41 +411,87 @@ impl VM {
...
-                    let v = self.stack.pop();
-                    let parent_closure = self.current_frame().closure(closure_depth);
-                    for (frame_depth, pframe) in self.frames.iter().rev().enumerate() {
-                        if Gc::ptr_eq(&parent_closure, &pframe.closure) {
-                            let sp = pframe.sp();
-                            self.stack.truncate(sp);
-                            self.stack.push(v);
-                            return SendReturn::ClosureReturn(frame_depth);
-                        }
+                    //XXX we aren't checking for escaped blocks!
```
vext01, line 475: If I were to nitpick, I'd say we need a space before `XXX`. Up to you.

src/lib/vm/somstack.rs
```diff
@@ -1,112 +1,157 @@
...
 pub const SOM_STACK_LEN: usize = 4096;
 
-/// A fixed size stack of SOM values. This stack does minimal or no checking on important
+/// A contiguous stack of SOM values. This stack does minimal or no checking on important
 /// operations and users must ensure that they obey the constraints on each function herein, or
 /// undefined behaviour will occur.
+///
+/// The basic layout of the stack is as a series of function frames growing from the beginning of
+/// the stack upwards. On a 64-bit machine, a function frame looks like:
+///
```
vext01, line 15: I'm struggling with this layout diagram.

I think we should define `n` and `m`. I think `n` is the number of arguments, what is `m`?

I *think* (but confirmation would be good) that `arg0 == var0` which is why we jump to `var n+1` after the args.

So perhaps `m` is the number of variables which are not arguments?

vext01:
I get the overall gist of this, but I won't to pretend that I understand the finer details.

I've done a pretty shallow review. I hope that's ok.

src/lib/compiler/ast_to_instrs.rs
ltratt: I think it's fairly obvious myself.

src/lib/vm/core.rs
ltratt: Oops, that comment is totally redundant now (I meant to remove it). Killed in https://github.com/softdevteam/yksom/pull/184/commits/5480fca4a7ba34caf27bb78943a867ad81f87b69.

src/lib/vm/somstack.rs
ltratt: Simplified in https://github.com/softdevteam/yksom/pull/184/commits/fcdd2d7199bc79800edbb8c279a30a8b793ca245.

ltratt:
I know this is borderline unreviewable, so shallow is fine! If you're OK with those changes, I can squash?

vext01:
Please squash.

ltratt:
Squashed.

PR #185, opened 2020-09-06
Cache a function's blocks

ltratt:
In functions with tight loops, continually allocating `Block`s (and `UpVar`s) is expensive. It's also unnecessary. Consider this contrived function:

```
    f = ( | x |
        x := 5.
        [x = 0] whileFalse: [
            (x > 2) ifTrue: [x println].
            x := x - 1.
        ].
    )
```

Each time `ifTrue` executes the inner block, it captures the same variables from the outer scope and operates on the same instance: there's no need to continually create a fresh block. This PR thus caches blocks, storing them on the SOM stack just after a function's variables.

How effective this is will depend hugely on a benchmark (a variant of the above program can easily see 30-40% improvement), but it seems to be roughly a 5-10% improvement on realistic benchmarks.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -367,7 +367,7 @@ impl VM {
         }
 
         let stack_base = self.stack.len() - func.num_params();
-        for _ in 0..func.num_vars() - func.num_params() {
+        for _ in 0..func.num_vars() - func.num_params() + func.num_blocks() {
             self.stack.push(self.nil);
         }
         // Inside this function, one should use this macro instead of calling `send_args_on_stack`
@@ -410,6 +410,13 @@ impl VM {
                     pc += 1;
                 }
                 Instr::Block(blkfn_idx) => {
+                    let cnd = self.stack.peek_at(stack_base + func.num_vars() + blkfn_idx);
+                    if cnd != self.nil {
+                        let blk = cnd.downcast::<Block>(self).unwrap();
+                        self.stack.push(cnd);
+                        pc = blk.func.bytecode_end();
+                        continue;
+                    }
                     let blk_func = func.block_func(blkfn_idx);
                     let mut blk_upvars = Vec::with_capacity(blk_func.upvar_defs().len());
                     for u in blk_func.upvar_defs() {
@@ -462,6 +469,7 @@ impl VM {
                         stack_base
                     };
                     let v = Block::new(self, rcv, blk_func, blk_upvars, sb);
+                    self.stack.set(stack_base + func.num_vars() + blkfn_idx, v);
                     self.stack.push(v);
                     pc = blk_func.bytecode_end();
                 }
@@ -592,7 +600,8 @@ impl VM {
                     };
 
                     if meth.func.is_primitive() && meth.func.primitive() == Primitive::Restart {
-                        self.stack.truncate(stack_base + func.num_vars());
+                        self.stack
+                            .truncate(stack_base + func.num_vars() + func.num_blocks());
                         pc = func.bytecode_off();
                         continue;
                     }
--- a/src/lib/vm/function.rs
+++ b/src/lib/vm/function.rs
@@ -152,6 +152,10 @@ impl Function {
         self.max_stack
     }
 
+    pub fn num_blocks(&self) -> usize {
+        self.block_funcs.len()
+    }
+
     pub fn block_func(&self, idx: usize) -> Gc<Function> {
         debug_assert!(idx < self.block_funcs.len());
         *unsafe { self.block_funcs.get_unchecked(idx) }
--- a/src/lib/vm/somstack.rs
+++ b/src/lib/vm/somstack.rs
@@ -19,7 +19,10 @@ pub const SOM_STACK_LEN: usize = 4096;
 ///   n * 8:     <var 0>
 ///              ...
 ///              <var m>
-///   (n+m) * 8: <working stack>
+///   (n+m)*8:   <block 0>
+///              ...
+///              <block p>
+///   (n+m+p)*8: <working stack>
 ///
 /// The compiler and VM treat <arg 0> as special: it always contains a reference to `self` (hence
 /// all functions have 1 extra parameter over those specified by the user). Functions are expected
@@ -27,7 +30,8 @@ pub const SOM_STACK_LEN: usize = 4096;
 /// to after the arguments but before the variables (i.e. the function will then set up its
 /// variables however it wants). Similarly, when a function is returned, the return value is
 /// expected to be placed where <arg 0> was originally found (i.e. at the end of the previous
-/// function's working stack).
+/// function's working stack). A functions `Block`s are cached after the variables, saving us from
+/// repeatedly allocating `UpVar`s for blocks in tight loops.
 pub struct SOMStack {
     /// How many items are used? Note that the stack has an implicit capacity of [`SOM_STACK_LEN`].
     len: usize,
--- a/src/lib/vm/somstack.rs
+++ b/src/lib/vm/somstack.rs
@@ -72,8 +72,9 @@ impl SOMStack {
         SOM_STACK_LEN - self.len()
     }
 
-    pub unsafe fn addr_of(self: Gc<Self>, n: usize) -> Gc<Val> {
-        Gc::from_raw(storage!(self).add(n))
+    /// Return the address of the variable `off` items from the beginning of the stack.
+    pub unsafe fn addr_of(self: Gc<Self>, off: usize) -> Gc<Val> {
+        Gc::from_raw(storage!(self).add(off))
     }
 
     /// Returns the top-most value of the stack without removing it. If the stack is empty, calling
@@ -82,7 +83,7 @@ impl SOMStack {
         self.peek_n(0)
     }
 
-    /// Peeks at a value `n` items from the top of the stack.
+    /// Peeks at a value `n` items from the beginning of the stack.
     pub fn peek_at(self: Gc<Self>, off: usize) -> Val {
         debug_assert!(off < self.len());
         unsafe { ptr::read(storage!(self).add(off)) }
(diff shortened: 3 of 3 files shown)


smarr:
You can cheat, as long as you don't get caught.

Try this one:

```
Hello = (
  | field |
  
  takeBlock: aBlock = (
    field == aBlock ifTrue: [ self error: 'ugh, this aint looking proper' ].
    field := aBlock.
    field value
  )
  
  run = (
    1 to: 10 do: [:i |
      self takeBlock: ['hurray I am a block, with my own identity' println] ]
  )
)
```

vext01:
bors try

ltratt:
@smarr That program gives the same result on Java SOM and yksom :) The only way you could detect something is different is, I think, something like:

```
arr := [].
1 to: 10 do: [:i arr.push [ <blah > ]]
```

and then to check by identity whether the blocks pushed into `arr` are distinct or not.

ltratt:
Ha, actually that test program won't do what I thought: it will fill it up with distinct blocks. So it is quite difficult to detect the caching!

smarr:
Hm, perhaps, I misunderstood what you are doing.

I would have expected your interpreter to fail early.

Is what you are doing not similar to this?

```
Hello = (
  | field |
  
  takeBlock: aBlock = (
    field == aBlock ifTrue: [ self error: 'ugh, this aint looking proper' ].
    field := aBlock.
    field value
  )
  
  run = (
    | block |
    1 to: 10 do: [:i |
      self takeBlock: ['hurray I am a block, with my own identity' println] ].
    
    block := ['hurray I am a block, with my own identity' println].
    1 to: 10 do: [:i |
      self takeBlock: block ].
  )
)
```

Which produces on an interpreter where each `[]` expression generates a distinct identity, the following:

```
./som-bc-interp -cp Smalltalk Examples/Hello.som
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity

ERROR: ugh, this aint looking proper
```

ltratt:
@smarr Calling that newer test program `Hello2.som`, on Java SOM:

```
$ zsh som.sh -cp Smalltalk /tmp/Hello2.som
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity

ERROR: ugh, this aint looking proper
```

On yksom with this branch:

```
$ cargo run -- --cp SOM/Smalltalk /tmp/Hello2.som
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/yksom --cp SOM/Smalltalk /tmp/Hello2.som`
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity
hurray I am a block, with my own identity

ERROR: ugh, this aint looking proper
Traceback (most recent call at bottom):
  File SOM/Smalltalk/System.som, line 46, column 8:
        (application respondsTo: #run:)
              ifTrue:  [ application run: arguments ]
              ifFalse: [ application run ]
  File SOM/Smalltalk/Boolean.som, line 31, column 8:
        self ifFalse: [ ^falseBlock value ].
  File SOM/Smalltalk/False.som, line 33, column 24:
    ifFalse: block = ( ^block value )
  File SOM/Smalltalk/Boolean.som, line 31, column 25:
        self ifFalse: [ ^falseBlock value ].
  File SOM/Smalltalk/System.som, line 48, column 23:
            ifFalse: [ application run ]
  File /tmp/Hello2.som, line 16, column 4:
    1 to: 10 do: [:i |
        self takeBlock: block ].
  File SOM/Smalltalk/Integer.som, line 65, column 8:
        self to: limit by: 1 do: block
  File SOM/Smalltalk/Integer.som, line 71, column 8:
        [ i <= limit ] whileTrue: [ block value: i. i := i + step ]
  File SOM/Smalltalk/Block.som, line 40, column 8:
        block value.
  File SOM/Smalltalk/Integer.som, line 71, column 36:
        [ i <= limit ] whileTrue: [ block value: i. i := i + step ]
  File /tmp/Hello2.som, line 17, column 6:
      self takeBlock: block ].
  File /tmp/Hello2.som, line 5, column 4:
    field == aBlock ifTrue: [ self error: 'ugh, this aint looking proper' ].
  File SOM/Smalltalk/True.som, line 32, column 24:
    ifTrue:  block = ( ^block value )
  File /tmp/Hello2.som, line 5, column 30:
    field == aBlock ifTrue: [ self error: 'ugh, this aint looking proper' ].
  File SOM/Smalltalk/Object.som, line 64, column 64:
    error: string = ( '' println. ('ERROR: ' + string) println. system exit: 1 )
exit(1).
```

so equivalent apart from the backtrace after `exit`.

Having investigated SOM a little more, I'm starting to think this optimisation only affects `Block::whileTrue:block`:

```
    whileTrue: block = (
        self value ifFalse: [ ^nil ].
        block value.
        self restart
    )
```

because that's the only place where a block is reexcuted in a single function frame. I sort-of forgot how weird looping is in SOM compared to other languages!

So an interesting question is whether `restart` is required to create completely new blocks with distinct identities or not.


Laurie

ltratt:
Looked at another way: this is, accidentally, an incredibly focussed performance hack to make up for our currently-slow GC. Given that it can affect at most one method, I need to think about whether a simple version of this is more appropriate.

smarr:
Hmm, I am very confused.

I don't see how the optimization is related to `while`, and how it would be restricted to it.

I don't think I properly understand the code:

https://github.com/softdevteam/yksom/pull/185/files#diff-1cfd31913e6ab1527c80b16531dba2e8R413-R414

```rust
let cnd = self.stack.peek_at(stack_base + func.num_vars() + blkfn_idx);
if cnd != self.nil {
```

So, you look at some point on the stack, and if that element isn't `nil`, then you apply your optimization.

How does this work and relate to restart?

If we look at `whileTrue:`, we see the following:

```smalltalk
    whileTrue: block = (
        self value ifFalse: [ ^nil ].
        block value.
        self restart
    )
```

So, something like `[ 'blockA' ] whileTrue: [ 'blockB' ]` involves three block objects.

The first block evaluates to `'blockA'`, the second to `'blockB'`, and the third possibly does a non-local return to exit `whileTrue` with `nil`.

The first two blocks are instantiated before the call.
And the `[ ^ nil ]` is instantiated for every iteration.

Going back to your earlier example, 

```smalltalk
[x = 0] whileFalse: [
  x > 2 ifTrue: [x println].
  x := x - 1.
].
```

Every time around the block, we got another one the `[x println]` block.
And of course, what ever happens in `whileFalse:`.

So, whileFalse looks like:

```smalltalk
    whileFalse: block = ( 
        [ self value not ] whileTrue: block 
    )
```

So, there's another block: `[ self value not ]`

And yes, this is expensive, if actually faithfully executing the standard library like this.
[Spoiler: Most SOMs don't because the standard library is declared as immutable, and therefore "lowering" it into the interpreter is not observable.

Anyway, I still don't see why the optimization triggers, but, it feels like the following should lead to an observable problem, because, structurally, it should match your example. Well, actually, it also doesn't. You may need to flip the condition and use `whileFalse` instead. But I would assume this breaks:

```Smalltalk
Hello = (
  | field |
  
  takeBlock: aBlock = (
    field == aBlock ifTrue: [ self error: 'ugh, this aint looking proper' ].
    field := aBlock.
    field value
  )
  
  run = (
    | block i |
    i := 1.
    [i < 5] whileTrue: [
      self takeBlock: ['hurray I am a block, with my own identity' println].
      i := i + 1. ].
    
    block := ['hurray I am a block, with my own identity' println].
    [i < 10] whileTrue: [
      self takeBlock: block.
      i := i + 1. ].
  )
)
```

The "standard" solution is to implement interesting loop constructs as primitives in the interpreter, and make loops that way much more efficient. In standard SOM, the semantics are simply defined layered on top of `whileTrue` and `restart`.

From my perspective, there's absolutely nothing "weird" about it.
Instead, it's perfectly good practice. You only have a single "primitive" construct defining the language semantics of relevance, and the rest can be built from that primitive, from that single truth. Arguably, a good principle. Compilers use the same idea, and "desugar" various loop constructs to a common lower-level representation, too.

ltratt:
The optimisation triggers for this line:

```
self value ifFalse: [ ^nil ].
```

because otherwise a new block (that just returns `nil`) is created every time `restart` restarts the method. I'm not even sure what `restart`s precise semantics are intended to be: should it reset variables, for example?

smarr:
Since `restart` is not supposed to be used anywhere else than in `whileTrue:`, you only need to reset the program counter.
At least that's what the other implementations do, if they actually use `reset`.

Is the trigger condition something that only ever will be true in `whileTrue:`? Seems a little brittle, but I don't know the underlying design decision that my support it.
I'd probably go for a more robust replacement of all of `whileTrue:` in a primitive set on the block class on load.

For instance like here: https://github.com/smarr/RPySOM/blob/master/src/som/primitives/bc/block_primitives.py#L71

ltratt:
As a point of principle, yksom doesn't make any assumptions about the standard library other than the primitives: so if the SOM library gains a `whileTrue` primitive, so will yksom, but not until then. If I loosen that principle, I think I'll be implementing a variant of SOM (closely related to SOM, but not quite the same language). And, frankly, I don't really know enough about the language to know what a plausible variant will be, so I'll stick with vanilla SOM!

vext01:
In light of the above discussion, do we still want to go ahead with this?

If so, I'll do a review.

ltratt:
Let's pause this for the time being until I've thought through things a bit more.

vext01:
I'm going to unassign this for now to clear it from my assigned list.

Feel free to add me back later if you do need a review.

ltratt:
After 3.5 years I think we can assume we won't merge this, as it makes fair comparisons with other SOMs even harder than without it.

PR #186, opened 2020-09-08
Remove the Vec allocation in `Block`.

ltratt:
This is our now "standard" optimisation. However, it can, and does create, cycles which then causes Boehm to warn (endlessly...) that:

```
  GC Warning: Finalization cycle involving 0x7fa72b06aea0
```

In a sense this is really a problem for rboehm, because yksom doesn't define any `Drop` methods, and isn't in danger of violating the "don't call (directly or indirectly) anything with `Gc` in it in a `drop` method". However, until that's fixed, this commit is unmergeable.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -411,62 +411,63 @@ impl VM {
                 }
                 Instr::Block(blkfn_idx) => {
                     let blk_func = func.block_func(blkfn_idx);
-                    let mut blk_upvars = Vec::with_capacity(blk_func.upvar_defs().len());
-                    for u in blk_func.upvar_defs() {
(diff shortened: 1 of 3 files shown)


jacob-hughes:
It's a shame we have to revert this for now, but lets keep it as a branch in yksom so I can experiment with how the finalization semantics should play out in rboehm.

ltratt:
We don't want to merge this into master.

jacob-hughes:
Ahh, apologies, I misunderstood this as a revert commit.

ltratt:
@jacob-hughes We should probably have another think about this PR and what we should do with it.

ltratt:
This one's definitely passed its sell by date!

PR #190, opened 2020-09-30
Add primitives for Krun

jacob-hughes:
This is needed in order to properly benchmark yksom with Krun as it allows us to send the necessary metrics to Krun without tearing down the VM after each in-process execution.

Right now, this will fail to build without the user passing the necessary linker flags to link `libkrunruntime.so` via `RUSTFLAGS`. At the moment this is a draft PR until we best work out how to do this nicely. I think the ideal solution is to conditionally compile krun stuff behind a `benchmark-harness` feature.

@ltratt do you have any other suggestions? Perhaps we can avoid this altogether if SOM has FFI support? I couldn't see this though.


src/lib/vm/core.rs
```diff
@@ -912,6 +913,75 @@ impl VM {
...
+                            self,
+                            VMErrorKind::CantRepresentAsIsize,
+                        ))
+                    }
+                };
+                unsafe { krun::krun_measure(idx as libc::c_int) };
+                SendReturn::Val
+            }
+            Primitive::KrunGetCoreCyclesDouble => {
+                let core = match self.stack.pop().as_isize(self) {
```
jacob-hughes, line 938: Out of interest. Why does `as_isize` return an `Option` but `as_usize` return a `Result`? Is there a more idiomatic way to do this similar to using the `stry!` macro?

ltratt: Every non-big integer can be represented as an `isize`, but some non-big integers can't be represented as a `usize`. However, I think your implicit suggestion is correct: making both return `Result` wouldn't really hurt and might makes things a bit easier to understand.

ltratt:
This all looks good to me! I don't have much idea about the linker flags, but https://stackoverflow.com/a/50646443 suggests a possible answer (c.f. https://doc.rust-lang.org/cargo/reference/config.html)?

jacob-hughes:
I think that in general, a user would want to clone yksom into some wider Krun-like project. If we specify linker flags in the build script (or even attributes in source) then we to make assumptions about where to find `libkruntime.so` which I'm not sure is useful. I actually think that in this case it's acceptable to build and run yksom using `RUSTFLAGS` and appending to the `LD_LIBRARY_PATH` in the following way:

`RUSTFLAGS="-l kruntime -L /path/to/libkrun/" cargo build`


`LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/libkrun/ ./yksom --cp SOM/Smalltalk Program.som`

To make sure that we don't get linker errors when using yksom without krun, I think all this stuff should be feature gated. What do you think?

ltratt:
If it can be feature-gated, excellent! If it can't... maybe the right thing to do is fork yksom into `yksom_krun` so that it doesn't "pollute" the main cases?

smarr:
Just a note, and I have no intention to make things complicated for you.

I set up a yksom CI job for the SOM repository to see when changes to SOM would break yksom:
https://github.com/SOM-st/SOM/pull/57/commits/a9e66e6ed654737b543001dbe0620df18775fe58

I'd think, the practical benefit of that is likely not out-weighting making things more complex for you guys though.

src/lib/vm/core.rs
jacob-hughes: I'll raise a PR for this

ltratt: It's OK, I'll do it -- you have more important research to do :)

Cargo.toml
```diff
@@ -50,6 +51,7 @@ ryu = "1.0"
 smartstring = "0.2"
 static_assertions = "1.1"
 termion = "1.5"
+libc = "0.2"
```
ltratt, line 54: Is it possible to include this only when `krun_harness` is specified?

jacob-hughes:
Good point. Try this: 5b546df

ltratt:
Please squash.

jacob-hughes:
Squashed

jacob-hughes:
It seems the nightly alloc API has changed again, so this should be merge-able once https://github.com/softdevteam/rboehm/pull/28 has landed.

PR #192, opened 2020-10-08
Homogenise as_isize's return types

ltratt:
The main objective of this PR is to  make `Val::as_isize` return `Result<...>`.

This not only brings it in line with as_usize, but means that if the user makes a dynamic type error that passes a non-int to code that calls as_isize, they'll get a backtrace rather than an internal compiler error!

@jacob-hughes It might be interesting for you (and me!) to do a Krun run before and after this PR? The first commit (probably) reduces overhead while the second commit *might* introduce some overhead. [The second commit is arguably a correctness fit, so I wouldn't reject it for performance reasons; but it might still be interesting to know what the overall effect is!]

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -687,7 +687,7 @@ impl VM {
             }
             Primitive::As32BitSignedValue => {
                 let rcv = self.stack.pop();
-                let i = if let Some(i) = rcv.as_isize(self) {
+                let i = if let Ok(i) = rcv.as_isize(self) {
                     i as i32 as isize
                 } else {
                     rcv.downcast::<ArbInt>(self)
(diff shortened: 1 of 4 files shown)


jacob-hughes:
LGTM, I'll get back to you with the Krun results before we merge it.

ltratt:
Thanks!

jacob-hughes:
As per our benchmarking observations in Krun, it seems that this PR has little to no impact on performance. I'm happy with the code changes. Are you happy for me to merge this?

ltratt:
Works for me!

jacob-hughes:
bors try

jacob-hughes:
Not sure why I didn't r+ this one straight away...

bors r+

PR #195, opened 2020-10-29
Fix incorrect as_isize use inside Krun primitives

jacob-hughes:
This is conditionally compiled based on 'krun_harness', which is why it
went unnoticed by bors.


src/lib/vm/core.rs
```diff
@@ -927,38 +927,36 @@ impl VM {
             }
             #[cfg(feature = "krun_harness")]
             Primitive::KrunMeasure => {
-                let idx = match self.stack.pop().as_isize(self) {
-                    Some(i) => i,
-                    None => {
-                        return SendReturn::Err(VMError::new(
+                let idx = if let Ok(i) = self.stack.pop().as_isize(self) {
+                    i as i32 as isize
```
ltratt, line 931: Isn't `i` a `usize` anyway? I'm not sure why the `as i32 as isize` bit is needed?

ltratt: Ping.

jacob-hughes: Yeah I'm not sure it is. I sort of copied this from `src/lib/vm/core.rs:691`:

```rust
            Primitive::As32BitSignedValue => {
                let rcv = self.stack.pop();
                let i = if let Ok(i) = rcv.as_isize(self) {
                    i as i32 as isize
                } else {...
```

I assumed there might have been a good reason for it :)

ltratt: The clue might be in the Primitive::*As32Bit*SignedValue part of that bytecode! So there was a good reason for it there but not in your PR ;)

jacob-hughes: Ha good point, I was on autopilot here. Still, I'm not sure why in that case there's a `i as i32 as isize` dance either as `as_isize` already gives you an `isize`?

ltratt: How else can you guarantee that the value is a 32-bit value if `isize` is bigger than 64 bits? Basically the `as i32` chops off the high bits (depending on the platform).

jacob-hughes: No ignore me. I get it. Not really with it today...

jacob-hughes: Fixed 4e5792d57343815f4b6610dbf0ebe42d0c6968ea

ltratt:
Please squash.

jacob-hughes:
Squashed

jacob-hughes:
Fixed the formatting

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #196, opened 2020-11-16
We need to check BigInt::to_f64() for infinity explicitly.

ltratt:
Other SOMs that I've tried error (whether deliberately or not) when a big integer is too big to be represented as a floating point number. The `BigInt::to_f64()` function used to return `None` in such cases but now they always return `Some::(f64::INFINITY)` (see https://github.com/rust-num/num-bigint/pull/163). This PR handles that explicitly.

On the plus side, we can simplify our code a little bit; on the downside we'll also probably be a tiny bit slower in the common case.

### DIFF
--- a/src/lib/vm/objects/integers.rs
+++ b/src/lib/vm/objects/integers.rs
@@ -51,10 +51,7 @@ impl Obj for ArbInt {
         } else if let Some(rhs) = other.try_downcast::<ArbInt>(vm) {
             Ok(ArbInt::new(vm, &self.val + &rhs.val))
         } else if let Some(rhs) = other.try_downcast::<Double>(vm) {
-            match self.val.to_f64() {
-                Some(i) => Ok(Double::new(vm, i + rhs.double())),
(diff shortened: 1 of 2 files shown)


ltratt:
bors try

src/lib/vm/objects/integers.rs
```diff
@@ -388,6 +364,17 @@ impl ArbInt {
     pub fn bigint(&self) -> &BigInt {
         &self.val
     }
+
+    /// If this `ArbInt` can be represented as a finite `f64`, return that value; otherwise return
+    /// `Err(CantRepresentAsDouble`.
```
ptersilie, line 369: Missing `)` in the comment.

ltratt: Fixed in https://github.com/softdevteam/yksom/pull/196/commits/00dc726a0293a42e0db9038e2ec9c8190d08f8b5.

ptersilie:
Please squash.

ltratt:
Squashed.

ltratt:
@ptersilie I think we need to retry bors.

PR #197, opened 2020-11-18
Minimally add support for arbitrary sized integer constants.

ltratt:
This is needed to support a recent test added to the SOM test suite.

I could do a more sophisticated job here (such as caching the ArbInt objects), but this is good enough to pass the test suite, which is the first step.


ltratt:
bors try

ltratt:
Fixed in https://github.com/softdevteam/yksom/pull/197.

PR #198, opened 2020-11-19
yksom takes a while to build so allow up to 30 minutes.

ltratt:
Especially if the build server is performing other tasks at the same time, 10 minutes is starting to cut things a bit fine, and may explain the occasional bors timeouts we've seen recently.


ltratt:
bors try

PR #200, opened 2021-04-13
Fix yksom after libgc changes

jacob-hughes:
This updates various places in yksom to work after the following breaking changes to libgc:

- `Gc<T>` can no longer be mutably dereferenced, so the places in yksom which relied on this have been updated to use interior mutability instead.
- `Gc<T>` now requires `T: Send`, so the `Obj` and `Array` traits needed updating with this constraint also.

Please can you pay special attention to the changes I've made in `UpVars`. The first couple of times I'd tried this, I had broken the `block.som` and `escape{n}.som` tests. I seem to have got this working now, I'd got very confused over whether in places I needed to be taking the on-stack-reference, or the `usize` field inside `Val`.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -16,12 +16,6 @@ sh rustup.sh --default-host x86_64-unknown-linux-gnu \
     -y
 export PATH=`pwd`/.cargo/bin/:$PATH
 
-cargo test
-cargo test --release
-
-cargo run -- --cp SOM/Smalltalk SOM/TestSuite/TestHarness.som
-cargo run --release -- --cp SOM/Smalltalk SOM/TestSuite/TestHarness.som
-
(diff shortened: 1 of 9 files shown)


ltratt:
At first I wasn't sure about this, but having looked a bit more carefully, I now think it's OK. We could, perhaps, change the `Cell` in `SOMStack` to `AtomicUsize` but the practical effect would be zero. I think this is OK as-is.

bors r+

jacob-hughes:
Yes I think so. `Cell` doesn't implement `Sync`, so even though the `SOMStack` can be passed between threads, its data can't be accessed by another thread. An `AtomicUsize` (or `Mutex`) would work if we wanted its access to be synchronised, but it looks like it's only used by the main-thread at the moment.

ltratt:
yksom is single threaded so no problem there!

PR #201, opened 2021-04-15
Minimally update to the latest SOM library.

ltratt:
This commit is enough to pass the test harness, though it doesn't actually implement all the new primitives.


jacob-hughes:
LGTM

bors r+

smarr:
> it doesn't actually implement all the new primitives

Yeah, they are only used by SomSom. So, except if you want to climb the metacircular tower, you won't need them, I'd think.

PR #202, opened 2021-04-16
Two bug fixes

ltratt:
Both noticed when running SomSom, but we add to yksom's test suite for completeness.

### DIFF
--- a/src/lib/compiler/ast_to_instrs.rs
+++ b/src/lib/compiler/ast_to_instrs.rs
@@ -640,7 +640,16 @@ impl<'a, 'input> Compiler<'a, 'input> {
                     max_stack = max(max_stack, 1 + i + expr_stack);
                 }
                 let send_off = vm.add_send((mn, msglist.len()));
-                let instr = Instr::Send(send_off, vm.new_inline_cache());
+                let instr = match receiver {
+                    box ast::Expr::UnaryMsg {
+                        receiver: box ast::Expr::VarLookup(span2),
+                        ids,
+                        span: _,
+                    } if ids.len() == 0 && self.lexer.span_str(*span2) == "super" => {
+                        Instr::SuperSend(send_off, vm.new_inline_cache())
+                    }
+                    _ => Instr::Send(send_off, vm.new_inline_cache()),
+                };
                 vm.instrs_push(instr, *span);
                 debug_assert!(max_stack > 0);
                 Ok(max_stack)
--- a/lang_tests/keyword_super/superc.som
+++ b/lang_tests/keyword_super/superc.som
@@ -1,0 +1,5 @@
+superc = (
+    x: i = (
+        (i + 1) println.
+    )
+)
--- a/lang_tests/keyword_super/test.som
+++ b/lang_tests/keyword_super/test.som
@@ -1,0 +1,15 @@
+"
+VM:
+  status: success
+  stdout: 2
+"
+
+test = superc (
+    run = (
+        super x: 1.
+    )
+
+    x: i = (
+        i println.
+    )
+)
--- a/lang_tests/symbol.som
+++ b/lang_tests/symbol.som
@@ -12,6 +12,7 @@ VM:
     #ab
     #+
     #+-+
+    #primitive
 "
 
 symbol = (
@@ -29,5 +30,6 @@ symbol = (
 
         #+ println.
         #+-+ println.
+        #primitive println.
     )
 )
--- a/src/lib/compiler/som.y
+++ b/src/lib/compiler/som.y
@@ -175,6 +175,7 @@ StringConst -> Result<Expr, ()>:
       "#" "STRING" { Ok(Expr::StringSymbol(map_err($2)?.span())) }
     | "#" "ID" { Ok(Expr::Symbol(map_err($2)?.span())) }
     | "#" "KEYWORD" { Ok(Expr::Symbol(map_err($2)?.span())) }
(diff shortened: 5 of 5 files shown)


ltratt:
I've forgotten the pattern matching syntax! https://github.com/softdevteam/yksom/pull/202/commits/123e2d99ff3fceafbc3a2b2b011e8fd15354598a will need squashing.

src/lib/compiler/ast_to_instrs.rs
```diff
@@ -640,7 +640,16 @@ impl<'a, 'input> Compiler<'a, 'input> {
                     max_stack = max(max_stack, 1 + i + expr_stack);
                 }
                 let send_off = vm.add_send((mn, msglist.len()));
-                let instr = Instr::Send(send_off, vm.new_inline_cache());
+                let instr = match receiver {
+                    box ast::Expr::UnaryMsg {
+                        receiver: box ast::Expr::VarLookup(span2),
+                        ids,
+                        span: _,
+                    } if ids.len() == 0 && self.lexer.span_str(*span2) == "super" => {
```
smarr, line 648: This is a surprising change, but might be the way how this is structured.

Looking through the whole file, you special case this now in keyword and unary messages.
Does this also need to go into binary messages?

ltratt: I don't know what binary + `super` means to be honest.

The weirdness is due to this:

```
x = (
    run = (
        |y|
        y := super.
        y println.
    )
)
```

which in JavaSOM and yksom prints:

```
instance of x
```

In other words, AFAICS, `super` means something different when it's a receiver vs. a "normal" variable.

ltratt: I wasn't very clear about the implication of the above. It means that if you want `super` as a special type of receiver *and* as a variable you have a grammar that is not provably unambiguous (indeed, I'm pretty sure it's *really* ambiguous): I hacked a quick example and it leads to 1 shift/reduce and 21 reduce/reduce errors. That's why we have to do the disambiguation in an ugly fashion inside the compiler itself rather than in the parser.

smarr: Multiple things:

1. SOM doesn't have operators. It doesn't matter whether a message is unary, binary, or keyword. The lookup semantics are the same. Unary/binary/keyword just have different priorities/precedences.

You can define a `+` message on an arbitrary class.

2. reading `super` is equivalent to reading `self`.

3. sending a message to `super` means that the lookup starts at the class where the method is defined in which the super send is done, and then considers the methods from the superclass on.
A normal send starts always at the concrete class of the receiver.

ltratt: I'd happily take a suggestion for a test for a `super` send with binary operators to see if it does the right thing in yksom (it probably doesn't at the moment).

smarr: And yes, `super` has only a meaning when used in connection with a message send.

Assuming there are different definitions of `foobar` in the class hierarchy, the following can have different results:

```
| s |
super foobar println.  "this is a super send"
s := super.
s foobar println. "this is a normal send"
```

ltratt: Yes, `yksom` will treat `super foobar` differently from `s foobar` in the above.

ltratt: [Your example neatly shows the ambiguity surrounding `super` in SOM's grammar FWIW!]

smarr: For reference, section 3.4.7 of the ANSI standard seems to be what SOM intents to do for super: https://wiki.squeak.org/squeak/uploads/172/standard_v1_9-indexed.pdf I expressed that a little more awkwardly, I think.

ltratt: Interestingly that section says "[super] cannot be used in place of self as a value" -- though JavaSOM (at least) allows `super` to be used as a value (and yksom emulates that behaviour, hence the awkwardness you see above).

smarr: For a test, which TruffleSOM doesn't pass :sob: :

```Smalltalk
A = (
 + other = (
   'A +' println.
 )
 
 unary = (
   'A unary' println.
 )
 
 keyword: other = (
   'A keyword' println
 )
)
```

```Smalltalk
B = A (
  + other = (
    'B +' println.
  )
  
  unary = (
    'B unary' println.
  )
  
  keyword: other = (
    'B keyword' println
  )

  run = (
    self + 4.
    super + 5.
    
    self unary.
    super unary.
    
    self keyword: 3.
    super keyword: 4.
  )
)
```

Expected result:

```
./som -cp Smalltalk B
B +
A +
B unary
A unary
B keyword
A keyword
```

smarr: SOM (java) passed it. At least something. sigh.

Will add to the core-lib.

smarr: > Interestingly that section says "[super] cannot be used in place of self as a value" -- though JavaSOM (at least) allows super to be used as a value (and yksom emulates that behaviour, hence the awkwardness you see above).

This is likely the original sin in SOM, which has a bunch of awkward consequences: https://github.com/SOM-st/SOM/issues/6

jacob-hughes:
The changes inside the VM look sensible. Please squash.

ltratt:
Squashed.

PR #203, opened 2021-04-16
Fix binary super sends.

ltratt:
This allows yksom to pass https://github.com/SOM-st/SOM/pull/72. We also include a lang_test for completeness.


jacob-hughes:
LGTM

bors r+

PR #204, opened 2021-04-17
Get SomSom running

ltratt:
This PR is sufficient to get SomSom (a SOM interpreter written in SOM) up and running. Since this also serves as a large-scale test of yksom, I've added it to `.buildbot.sh`.

### DIFF
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -793,7 +793,9 @@ impl VM {
                 SendReturn::Val
             }
             Primitive::ErrorPrint => {
-                todo!();
+                let v = self.stack.pop();
+                eprint!("{}", stry!(v.downcast::<String_>(self)).as_str());
+                SendReturn::Val
             }
             Primitive::ErrorPrintln => {
                 let v = self.stack.pop();
--- a/src/lib/vm/core.rs
+++ b/src/lib/vm/core.rs
@@ -3,6 +3,7 @@
 use std::{
     collections::HashMap,
     convert::TryFrom,
+    fs::read_to_string,
     mem::size_of,
     path::{Path, PathBuf},
     process,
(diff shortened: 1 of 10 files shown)


ltratt:
bors try

ltratt:
Gah, different rustfmt versions again!

bors try

smarr:
Hm, nice. Looks like there were no major issues :)

ltratt:
bors try

ltratt:
@jacob-hughes This is ready for review (but will need squashing before merging).

src/lib/vm/core.rs
```diff
@@ -1048,7 +1049,17 @@ impl VM {
...
-                todo!();
+                let path_val = self.stack.pop();
+                let _rcv = self.stack.pop();
+                let path = PathBuf::from(stry!(path_val.downcast::<String_>(self)).as_str());
+                match read_to_string(path) {
+                    Ok(s) => {
+                        let v = String_::new_str(self, SmartString::from(s));
+                        self.stack.push(v);
+                    }
+                    Err(_) => self.stack.push(self.nil)
```
jacob-hughes: Is this right? It looks like if SOM can't load a file it will just return `Nil`? I would expect it to crash or throw an exception or something.

smarr: Returning `nil` on error is what the current primitive definition is.
Crashing is really not an option, because SOM uses classpath probing, and this here is the minimal interface to be able to try to read a file.

SOM doesn't have exceptions, so that's not an option either.
A more complete API would provide a `#loadFile:onError:` primitive, where the `onError:` part takes a block, which possibly gets information about the error to then handle it.

jacob-hughes: Ah, that makes sense. Thanks :)

jacob-hughes:
LGTM, please squash

src/lib/vm/core.rs
ltratt: I originally implemented this as throwing an exception before realising that it's really a "try loading a file" thing. I should have checked first -- I'd have saved myself some effort!

ltratt:
Squashed.

src/lib/vm/core.rs
smarr: In a slightly more perfect world, I would have done the better thing and implemented this:

```Smalltalk
System = (
  "..."
  loadFile: name onError: aBlock = primitive
 
  loadFile: name = (
    ^ self loadFile: name onError: [:err | ^ nil ]
  )
)
```

But, I didn't think further than SomSom...
So, yeah, the `#loadFile:` is what we got at the moment.

PR #206, opened 2021-04-21
Null entries on the SOMStack after they're removed.

ltratt:
This prevents the GC thinking they're still alive. It would be preferable to tell the GC "only scan the first N words of the stack" but we don't have any way of doing that at the moment AFAIK.

This doesn't have any performance difference that I can measure either way.

### DIFF
--- a/src/lib/vm/somstack.rs
+++ b/src/lib/vm/somstack.rs
@@ -156,7 +156,8 @@ impl SOMStack {
         debug_assert!(len <= self.len());
         for i in len..self.len() {
             unsafe {
-                ptr::read(storage!(self).add(i));
+                // Since `Val`s don't have a `drop` implementation, we can simply discard them
+                // without worrying about them being dropped.
                 ptr::write(storage!(self).add(i), Val::illegal());
             }
         }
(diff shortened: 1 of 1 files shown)


jacob-hughes:
> It would be preferable to tell the GC "only scan the first N words of the stack" but we don't have any way of doing that at the moment AFAIK

There is an API for this in Boehm, but I'm not sure how we'd use this with SOMStack. Maybe something like `new_from_layout_with_scan_limit(layout: Layout, limit: usize)`. 

These changes LGTM though.

bors r+

ltratt:
> There is an API for this in Boehm, but I'm not sure how we'd use this with SOMStack. Maybe something like new_from_layout_with_scan_limit(layout: Layout, limit: usize).

It's not a limit that you know at creation time: it's dynamic. So we'd have to have some way of either reading a `usize` or calling a user function to say "tell me how much of this chunk of memory is actually in use".

[FWIW, I am now wondering what happens when you do `pop` on a Rust `Vec`... If it doesn't null out the deleted entry -- which it might not do so as there's no correctness reason for it to do so -- we'll get memory leaks.]

PR #210, opened 2021-05-06
Ensure yksom works safely with untraceable allocation

jacob-hughes:
Some types (e.g. `Val`) have opaque pointers which are hidden from type system. This ensures that the GC can see them and still treats them as potential pointers during marking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 Cargo.lock
 target/
 *.core
+rebench.conf
+rebench.data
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 Cargo.lock
 target/
 *.core
+rebench.conf
```
smarr, line 4: One could argue that this one should be part of the repo, and part of CI ;)

PR #214, opened 2022-09-26
Use Natrob with rustgc feature flag

jacob-hughes:


ltratt:
I think `Cargo.toml` needs updating too, as it references `libgc`? Let's try.

bors try

jacob-hughes:
Oops I forgot the Cargo.toml. I can squash when you're ready

ltratt:
I think you need to push :)

ltratt:
bors try-

jacob-hughes:
Pushed, happy for me to squash?

ltratt:
Please squash.

jacob-hughes:
Squashed

ltratt:
bors try

vext01:
bors try-

vext01:
bors try-

vext01:
bors try

ltratt:
It's got past the git stage now, which is good!

bors try-

ltratt:
bors ping

ltratt:
bors try-

vext01:
bors try

ltratt:
@jacob-hughes We have an actual error for you to fix (I think) after all the noise!

ltratt:
@jacob-hughes Should we expect this PR to succeed now?

jacob-hughes:
bors try

src/lib/vm/objects/instance.rs
```diff
@@ -17,8 +17,6 @@ pub struct Inst {
     inst_vars: SyncUnsafeCell<Vec<Val>>,
 }
 
-unsafe impl std::gc::NoFinalize for Inst {}
-
```
jacob-hughes, line 21: Argh! I got confused and removed this by accident. I'll push a fix where I add it again.

jacob-hughes: Fixed: b4e0b75bd476a7cce1761d2b014c45aef1fa83b0

ltratt:
bors try

src/lib/vm/objects/class.rs
```diff
@@ -38,6 +38,10 @@ pub struct Class {
     inst_vars: UnsafeCell<Vec<Val>>,
 }
 
+// This is safe because the drop methods which run from fields in `Class` only
+// do deallocation.
+unsafe impl FinalizerSafe for Class {}
```
ltratt, line 43: Hmm, I wonder if this is safe? `PathBuf` and `HashMap` (two of them!) aren't obviously `FinalizerSafe` to me.

src/lib/vm/objects/string_.rs
```diff
@@ -32,6 +32,12 @@ pub struct String_ {
 #[cfg(feature = "rustgc")]
 impl NoFinalize for String_ {}
 
+// By default, `String_` is not `FinalizerSafe` because it contains a `Cell`
+// field. Cell is `!Sync` so it would be unsound to access inside a `drop`
+// method. This explicit impl is needed to tell the compiler we are not doing
+// that.
+unsafe impl FinalizerSafe for String_ {}
```
ltratt, line 39: I'm not sure about this as there's an underlying `SmartString` that looks like the drop method really does need to run sometimes https://github.com/bodil/smartstring/blob/e407ca23c747257a812d2d2e70bf336412718c3a/src/lib.rs#L196

jacob-hughes: You're right, this comment is not the full story. This _is_ safe because the drop method for `SmartString` won't be run while some other thread holds a reference to it (you'd need an `Arc` or equivalent for that). This isn't explained properly though so I'll change this.

src/lib/vm/objects/class.rs
jacob-hughes: A bit like `String_`, this is safe because `PathBuf` and `HashMap` won't be called while some other thread has a reference to them via an `Arc` or some other shared ownership pointer.

jacob-hughes: To put another way, we know (because we wrote yksom) that although `drop` will run on another thread, it will always happen when the object is dead and inaccessible from any other thread.

ltratt: True!

jacob-hughes: I guess this is the real reason. Should I make this the comment instead?

ltratt: Yes, I think that's a good idea (in retrospect the comment nudged me in the direction of confusing `FinalizerSafe` with `NoFinalize`; I'm an idiot and took the nudge!).

jacob-hughes: Easily done!

jacob-hughes: I've had a go at rewording these comments here: 7f662ffc4642d32b2d84c7f613a715d9d1e8f39f

jacob-hughes: ...It helps if I push them

ltratt:
I would say "please squash" but I note that the previous `bors try` failed. Does it need investigating?

jacob-hughes:
I think I fixed the problem here [eea626c](https://github.com/softdevteam/yksom/pull/214/commits/eea626cc1a119552d32597901b2125d076112cfb), so it should be ready to squash.

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #216, opened 2022-11-21
Switch to use FinalizerOptional trait

jacob-hughes:
This one requires https://github.com/softdevteam/natrob/pull/41 to land first.


jacob-hughes:
@ltratt It should be possibly to merge this one now.

PR #218, opened 2022-12-22
Fix integer multiplication.

ltratt:
Previously we didn't detect some cases of overflow properly due to integer tagging.

Spotted indirectly by @smarr in #217.


jacob-hughes:
LGTM

bors r+

PR #220, opened 2023-01-16
Run a benchmark in CI to make sure that we catch changes to the runner.

ltratt:
https://github.com/SOM-st/SOM/pull/82 changed how the benchmark runner works, but we didn't notice, and confused ourselves! With luck, this addition might spot such problems in the future.

[CC @smarr -- and thanks!]

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -34,7 +34,10 @@ cargo +alloy test --release
 cargo +alloy run  -- --cp SOM/Smalltalk SOM/TestSuite/TestHarness.som
 cargo +alloy run --release -- --cp SOM/Smalltalk SOM/TestSuite/TestHarness.som
 
-cd SOM && \
-  cargo +alloy run --release -- \
-    --cp Smalltalk:TestSuite:SomSom/src/compiler:SomSom/src/vm:SomSom/src/vmobjects:SomSom/src/interpreter:SomSom/src/primitives \
-    SomSom/tests/SomSomTests.som
+cd SOM
+cargo +alloy run --release -- \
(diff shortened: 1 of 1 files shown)


smarr:
For at least SOM-java, I do this:

https://github.com/SOM-st/som-java/blob/master/.github/workflows/ci.yml#L39-L50
```yaml
      - name: Checkout AWFY
        uses: actions/checkout@v3
        with:
          repository: smarr/are-we-fast-yet
          path: are-we-fast-yet

      - name: AWFY Test Run
        run: |
          pwd
          export JAVA_HOME=$JAVA_HOME_8_X64
          pip install ReBench
          rebench awfy-test.conf
```

https://github.com/SOM-st/som-java/blob/master/awfy-test.conf

Mostly, because I don't really run the AWFY benchmarks in that version else where, and still haven't managed to integrated them into the SOM repo.
And because test coverage isn't 100%...

ltratt:
Interesting! We should perhaps do something a bit like that at some point, but at least this current PR is better than nothing (I hope...).

jacob-hughes:
This can be merged once this lands: https://github.com/softdevteam/yksom/pull/221

ltratt:
@jacob-hughes I think this one's ready now.

ltratt:
bors try

ltratt:
bors try

ltratt:
`thread 'main' has overflowed its stack`

That's a new one! Let's see if it happens again.

bors try

ltratt:
I messed up the merge commit (`git add .` was my downfall). f02c837 should fix things.

bors try

jacob-hughes:
Please squash if it builds

ltratt:
Squashed.

PR #222, opened 2023-01-19
Update dependencies

ltratt:
This is a long overdue update of yksom's dependencies. Some are trivial, some (well... grmtools!) aren't. I've split them as separate commits so that if we find a problem later (I'm hopeful we won't but...), bisecting will be easier.

Note that I have also (belatedly) added a `Cargo.lock` (hence the bonkers looking number of added lines!).


jacob-hughes (approved):
LGTM

bors r+

Issue #224, opened 2023-01-23
[Feature Request] Improved error message on file not found, and support to give class name instead of file


smarr:
[Very Very Low Priority, but I thought I mention it]

At the moment, when a command line such as the following is given: `-cp Smalltalk Harness List`, yksom will bark with the following error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/main.rs:59:64
```

It would be great if the error could be a little more user friendly, and at least name which file wasn't found. Ideally, it would also avoid the Rust-specific output.

In the specific example, the problem is that `Harness` is actually `Harness.som` on the disk, and thus, not found.

The second feature request would be to support giving the class name for an entry point/main script, in addition to supporting the use of file names.

All other SOM implementations will try to load `Harness` from the classpath.
I didn't look up the exact logic implemented, but this works across SOMs:

```
SOMpp$ ./SOM++ -cp Smalltalk Examples/Hello.som
This is SOM++
Hello, World from SOM

SOMpp$ ./SOM++ -cp Smalltalk:Examples Hello
This is SOM++
Hello, World from SOM
```

ltratt:
This should be fixed by #226 -- comments welcome!

PR #226, opened 2023-01-24
More standard command line

ltratt:
Fixes https://github.com/softdevteam/yksom/issues/224.

@smarr Does this do things right? I've done a bit of guessing here...

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,12 +30,13 @@ fn usage(prog: &str) -> ! {
         .file_name()
         .map(|x| x.to_str().unwrap_or("yksom"))
         .unwrap_or("yksom");
-    writeln!(
-        &mut stderr(),
-        "Usage: {} [-h] --cp <path> <file.som> [-- <arg_1> [... <arg_n>]]",
-        leaf
-    )
-    .ok();
+    error(&format!(
+        "Usage: {leaf} [-h] --cp <path> <file.som> [-- <arg_1> [... <arg_n>]]"
+    ));
+}
+
+fn error(msg: &str) -> ! {
+    stderr().write_all(msg.as_bytes()).ok();
     process::exit(1)
 }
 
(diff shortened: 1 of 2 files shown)


ltratt:
@jacob-hughes I think we should go ahead with this one: I think it is an improvement (even if it might be possibly be improved even further).

src/main.rs
```diff
@@ -30,12 +30,13 @@ fn usage(prog: &str) -> ! {
...
         .map(|x| x.to_str().unwrap_or("yksom"))
         .unwrap_or("yksom");
-    writeln!(
-        &mut stderr(),
-        "Usage: {} [-h] --cp <path> <file.som> [-- <arg_1> [... <arg_n>]]",
-        leaf
-    )
-    .ok();
+    error(&format!(
+        "Usage: {leaf} [-h] --cp <path> <file.som> [-- <arg_1> [... <arg_n>]]"
```
jacob-hughes, line 34: Stupid question: what does leaf mean?

src/main.rs
```diff
@@ -56,18 +57,39 @@ fn main() {
...
-        None => vec![],
+    let mut cp: Vec<_> = matches
+        .opt_str("cp")
+        .unwrap()
+        .split(":")
+        .map(|x| PathBuf::from_str(x).unwrap())
+        .collect();
+
+    let mut src_path = Path::new(&matches.free[0]).to_owned();
+    if !src_path.is_file() && src_path.extension().is_none() {
```
jacob-hughes, line 68: Would it be more readable here to use `src_path.is_dir()`?

jacob-hughes (commented):
I like these changes. Looks mostly good, a couple of minor comments.

src/main.rs
ltratt: It's the name of the binary. yksom *tries* to lookup what its binary name is; when it fails (and it can fail, for weird reasons) it falls back on assuming it's called `yksom`.

src/main.rs
ltratt: Not only more readable -- but more correct! Fixed in 39b9452.

ltratt:
@jacob-hughes We might be ready for squashing.

src/main.rs
jacob-hughes: Does this now make the extra `.extension().is_none()` check redundant?

ltratt: Oh, hang on, I'm an idiot. This was right before, I think. The `!is_file()` was a combination of "there is nothing at this path and if there is something it mustn't be a file". I think I should revert 39f9452.

jacob-hughes: I see, ok yeah maybe best to revert then.

ltratt: Reverted.

jacob-hughes:
Please squash

ltratt:
Squashed.

PR #229, opened 2024-01-26
Increase the SOM stack size.

ltratt:


smarr:
I can confirm that Havlak executes with that stack size.

PR #231, opened 2024-10-23
Rename FinalizerOptional trait

jacob-hughes:


jacob-hughes:
Fixed in f47421dfcd96504696164dbbfe4520d1edbe5201, I don't think it should be squashed

jacob-hughes:
We need to use the latest version of debian to build LLVM in CI as it requires a more recent version of CMake. Fixed here (6f3d526d5f12d99a71857a003379321ad7d1f898)

jacob-hughes:
Fixed here ff35599b1f4983cbab7e7052fef8af4a2be26930

ltratt:
Please squash (if needed).

jacob-hughes:
Squashed

